PR #1133, opened 2021-03-12
fix: temporarily skip spurious test failure

digitalcora:
We have a task in the queue to address the various test failures caused by live data changing, and in the meantime this frequently blocks other tasks for no good reason.

### DIFF
--- a/apps/alert_processor/test/alert_processor/service_info/service_info_cache_test.exs
+++ b/apps/alert_processor/test/alert_processor/service_info/service_info_cache_test.exs
@@ -332,11 +332,9 @@ defmodule AlertProcessor.ServiceInfoCacheTest do
     assert {:ok, nil} = ServiceInfoCache.get_parent_stop_id(pid, "garbage")
   end
 
-  # Skipped: This function exists only to support ferry routes, and the test fails when the
-  # specified Charlestown ferry trip is not running on the current day. All ferry service is
-  # canceled as of 2020-03-16 due to COVID-19 service changes, but even during regular service,
-  # this fails when the current day is not a weekday.
-  @tag :skip
+  # TODO: address these tests having a dependency on live data that changes frequently
+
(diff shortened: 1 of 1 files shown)


apps/alert_processor/test/alert_processor/service_info/service_info_cache_test.exs
```diff
@@ -332,18 +332,17 @@ defmodule AlertProcessor.ServiceInfoCacheTest do
     assert {:ok, nil} = ServiceInfoCache.get_parent_stop_id(pid, "garbage")
   end
 
-  # Skipped: This function exists only to support ferry routes, and the test fails when the
-  # specified Charlestown ferry trip is not running on the current day. All ferry service is
-  # canceled as of 2020-03-16 due to COVID-19 service changes, but even during regular service,
-  # this fails when the current day is not a weekday.
-  @tag :skip
+  # TODO: address these tests having a dependency on live data that changes frequently
```
arkadyan, line 335: My standard reaction to seeing a `TODO` is to suggest that you might consider adding an Asana ticket for this instead in order to have better visibility and the ability to prioritize and link to existing tickets this is waiting on.

arkadyan (approved):
With a nudge for a ticket, this looks good.

apps/alert_processor/test/alert_processor/service_info/service_info_cache_test.exs
digitalcora: There is a task I alluded to in the description ([it's this one](https://app.asana.com/0/1167196461144361/1199513856779936)), though more for addressing this whole class of problems versus just these two cases. I can make sure that task mentions this specific place in the tests.

arkadyan: Oh, thanks, missed that. Yeah, whatever works best for you really.

PR #1142, opened 2021-04-05
Fix test failures and subscriptions page crash

digitalcora:
If a user had any subscriptions involving a route that didn't exist anymore, the subscriptions page would crash on render, preventing any further managing of subscriptions. This also blocked the test suite from passing, as the recently eliminated bus route 57A was used in a few test cases.

This fixes the immediate issue by displaying "Unknown Route" as the route name. For now attempting to edit such a subscription will still cause a crash, but it can be deleted, and will otherwise not cause any harm.

### DIFF
--- a/apps/concierge_site/lib/views/route_helper.ex
+++ b/apps/concierge_site/lib/views/route_helper.ex
@@ -6,7 +6,7 @@ defmodule ConciergeSite.RouteHelper do
   alias AlertProcessor.Model.{Route, Subscription}
 
   @doc """
-  Get the name of a given route or routes.
+  Get the name of a route, with a fallback name if the route isn't in the service info cache.
 
     iex> ConciergeSite.RouteHelper.route_name("Green")
     "Green Line"
@@ -20,6 +20,8 @@ defmodule ConciergeSite.RouteHelper do
     "Route 39"
     iex> ConciergeSite.RouteHelper.route_name("741")
     "Silver Line SL1"
+    iex> ConciergeSite.RouteHelper.route_name("no-such-route")
+    "Unknown Route"
   """
   @spec route_name(String.t()) :: String.t()
   def route_name(routes) when is_list(routes), do: "Green Line"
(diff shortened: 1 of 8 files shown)


apps/concierge_site/lib/views/route_helper.ex
```diff
@@ -30,13 +32,12 @@ defmodule ConciergeSite.RouteHelper do
...
-
-    # We never want long names for buses
-    if route.route_type == 3 || route.long_name == "" do
-      Route.bus_short_name(route)
-    else
-      route.long_name
+    case ServiceInfoCache.get_route(route_id) do
+      # We never want long names for buses
+      {:ok, %{route_type: 3} = route} -> Route.bus_short_name(route)
+      {:ok, %{long_name: ""} = route} -> Route.bus_short_name(route)
```
arkadyan, line 38: Codecov is calling out this line as not being covered by tests. Is it easy to add a case for this? I realize it's existing functionality not related to your changes, so it's fine if it's not trivial or doesn't seem like the time and place.

apps/concierge_site/lib/views/route_helper.ex
```diff
@@ -30,13 +32,12 @@ defmodule ConciergeSite.RouteHelper do
...
 
   def route_name(route_id) do
-    {:ok, route} = ServiceInfoCache.get_route(route_id)
-
-    # We never want long names for buses
-    if route.route_type == 3 || route.long_name == "" do
-      Route.bus_short_name(route)
-    else
-      route.long_name
+    case ServiceInfoCache.get_route(route_id) do
```
arkadyan, line 35: 👍

arkadyan (approved):
Looks good 👍

apps/concierge_site/lib/views/route_helper.ex
digitalcora: On closer inspection it seems like this was sort of duplicating logic that already existed in the form of `Route.name`, so I'm replacing it with a call to that. It's not exactly the same, but it passes all the current tests and I'm honestly not sure under what circumstance we'd want the "`bus_short_name`" behavior for routes that are not bus routes.

PR #1143, opened 2021-04-05
chore: update ehmon to work with newer OTPs

digitalcora:
When we previously upgraded to Erlang 22, it broke some graphs on our dashboard because Heroku's `ehmon` doesn't work on 21.2 and later. We made our own fork of the project that fixes this issue and are already using it on some other apps.


arkadyan (approved):

digitalcora:
Deployed this to dev-green and confirmed the dashboard graphs work again in that environment.

PR #1144, opened 2021-04-06
fix: alert processing crash in `get_route`

digitalcora:
In c4914b5 we removed a seemingly unused clause of `get_route` that allowed passing `nil` as the route ID. Though this was not covered by any tests, it turned out to be used when an uncommon set of conditions occurred, and this caused processing of the "oldest" group of alerts to always crash in production.

This adds two tests that both fail in the presence of this bug: one for `ServiceInfoCache` that explicitly tests passing `nil` as the route ID, and one for `InformedEntityFilter` replicating the conditions that cause `nil` to be passed as the route ID.

This also fixes an incorrect pattern in `InformedEntityFilter` that never would have matched, because it was matching a subscription `type` (which only has atom values) against a string.

### DIFF
--- a/apps/alert_processor/lib/rules_engine/informed_entity_filter.ex
+++ b/apps/alert_processor/lib/rules_engine/informed_entity_filter.ex
@@ -175,7 +175,7 @@ defmodule AlertProcessor.InformedEntityFilter do
     MapSet.subset?(origin_and_destination, scheduled_stops)
   end
 
-  defp in_between_stop_match?(%{type: "accessibility"}, _), do: false
+  defp in_between_stop_match?(%{type: :accessibility}, _), do: false
 
   defp in_between_stop_match?(subscription, %{activities: activities, stop: stop}) do
     stop in stops_in_between(subscription, activities)
--- a/apps/alert_processor/lib/service_info/service_info_cache.ex
+++ b/apps/alert_processor/lib/service_info/service_info_cache.ex
@@ -78,7 +78,7 @@ defmodule AlertProcessor.ServiceInfoCache do
   @spec get_route(
           GenServer.server(),
           String.t() | %{route_id: String.t(), stop_ids: [String.t()]}
-        ) :: {:ok, Route.t()} | {:error, any}
+        ) :: {:ok, Route.t()} | {:error, atom}
   def get_route(name \\ __MODULE__, route)
 
   def get_route(name, route) when is_binary(route) do
@@ -101,6 +101,8 @@ defmodule AlertProcessor.ServiceInfoCache do
   end
 
   def get_route(name, %{route_id: route}), do: get_route(name, route)
+
+  def get_route(_, _), do: {:error, :invalid_argument}
 
   def get_routes(name \\ __MODULE__) do
     GenServer.call(name, :get_routes, @timeout)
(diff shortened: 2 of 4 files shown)


arkadyan (commented):
Could you say more about the exiting stop subscriptions? My understading of the rules around all this is really rusty… My first guess is to wonder, "if I have a subscription to stop X without a route specified, does that mean that it _should_ match an alert for a specific route serving my stop (i.e. match any route)?" But maybe I'm thinking about that wrong.

Feel free to ping me on Slack if it's easier to talk through this in person.

digitalcora:
> "if I have a subscription to stop X without a route specified, does that mean that it _should_ match an alert for a specific route serving my stop (i.e. match any route)?"

In this case the alert is for a different stop than the one your subscription is for, so it should not match (note in the test the subscription is for `"some stop"` but the informed entity applies to `"some other stop"`).

The specific condition that triggers the bug is:

1. An alert exists with an informed entity that:
   a. applies to a specific route _and_ stop
   b. applies to the `RIDE` activity at that stop
2. A "stop subscription" exists for a stop that is _not_ the specific stop the entity applies to

The informed entity matching goes like this:

1. Trip matching doesn't apply since the entity does not specify a trip
2. Route type matching doesn't apply since neither the entity nor the subscription has a specific route type
3. Direction ID matching doesn't apply since neither the entity nor the subscription has a specific direction ID
4. Route matching doesn't apply since the route of a "stop subscription" is `nil` (this is considered to match all routes)
5. For stop matching:
   a. Since neither the `origin` of the subscription nor the `stop` of the entity are `nil`, we are not in any of the special cases of `stop_match?`
   b. Since neither the subscription's `origin` nor `destination` are the same as the informed entity's `stop`, we call into `in_between_stop_match?` to determine whether the entity's stop is between the subscription's origin and destination
   c. This is not an `accessibility` subscription *️⃣, so we are not in the special case of `in_between_stop_match?`
   d. We continue into `stops_in_between`, where:
      1. `RIDE` is among the entity's activities, so that match succeeds
      2. We construct a "route filter" that **includes the subscription's `route`, which is `nil`** ⚠️
      3. We pass that filter to `ServiceInfoCache.get_route`, which crashes 🛑

*️⃣ It might seem weird that a "stop subscription" could exist outside the context of accessibility subscriptions, since I had thought that was the only way to create one. What might have actually happened on production is that the offending subscription _was_ an accessibility one, but was not caught by the special case of `in_between_stop_match?` due to the string/atom confusion that this PR also fixes. However, the existing tests seem to assume a non-accessibility stop subscription can exist. This could simply be a case of the test data being "unrealistic" and the validations not forcing it to be realistic. In any case, our bases should now be covered.

arkadyan (approved):
Thanks for the extended explanation, that's very helpful. This all semes great to me.

I don't know if you've poked around at all, but do you have any sense of how many of this sort of "non-accessibility stop subscription" there are? Especially if it's a small number, maybe it's worth a future ticket to look into them, confirm your guess about them being a case of bad data, and cleaning them up in whatever way makes most sense?

digitalcora:
My guess is that the number is zero, the tests are just not using 100% realistic data, and the production crash was a result of the string/atom mixup in `in_between_stop_match?`. But yeah, might be a good thing to follow up on.

arkadyan:
@digitalcora Ok, yeah I guess I just made an assumption that maybe existing data was causing the crashes.

PR #1145, opened 2021-05-03
refactor: enable running on multiple instances

digitalcora:
This is a narrowly-targeted change to address our lack of production redundancy, and allow the app to be deployed without taking it offline for the duration. In particular, this does not enable alert processing to scale horizontally.

There are a few moving parts:

* `AlertProcessor.Lock` is introduced, a database-backed lock that can only be acquired by one instance at a time.

* `Model.Metadata` is introduced, a method of storing generic key-value data in the database.

* `AlertWorker` is rewritten to use `Lock` to ensure no other instances are processing alerts at the same time, and `Metadata` to determine when each duration type was last processed, even if it was processed by a different instance.

* `deploy.sh` is modified to just update the service with the new task definition, instead of taking it down, updating it, and bringing it back up. Most of this is borrowed from the API dev-blue deploy script.

* `prod.vm.args` is created for the Distillery release, allowing us to set the `sname` option. This causes a "hostname" like `ip-x-x-x-x` to be generated at runtime, which will then appear in the "node" part of log lines and allow us to differentiate between instances.

Since alert processing speed has improved quite a bit since the numbers were originally set, this also adjusts the interval for "oldest" alerts, which are now processed every 10 minutes instead of every 60. Typically this group only takes ~10 seconds to process on production.

### Deployment

This branch is currently deployed to dev-green, and dev-green has also been upgraded to 2 instances. Looking at the `AlertWorker` activity [in the Splunk logs](https://mbta.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22alerts-concierge-dev-green%22%20AlertWorker&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=-15m&latest=now&sid=1620064563.780136) you can see the exclusive lock at work, where occasionally something like this will happen (newest at the top):

```
2021-05-03T17:55:57.475 [info] node=alerts_concierge@ip-A AlertWorker event=check_end
2021-05-03T17:55:57.433 [info] node=alerts_concierge@ip-B AlertWorker event=check_end
2021-05-03T17:55:57.432 [info] node=alerts_concierge@ip-B AlertWorker event=skip reason=lock_in_use
2021-05-03T17:55:57.429 [info] node=alerts_concierge@ip-B AlertWorker event=check_start
2021-05-03T17:55:57.412 [info] node=alerts_concierge@ip-A AlertWorker event=process duration_type=recent seconds_outdated=2
2021-05-03T17:55:57.410 [info] node=alerts_concierge@ip-A AlertWorker event=check_start
```

If this code is deployed on an environment that still only has one instance (e.g. production), it will actually retain the old behavior of shutting down, updating the service, and starting up — it turns out ECS can do this just fine without us scripting it. So the intended flow is 1) deploy this, then 2) modify the environment to have two instances. This app synchronously runs migrations at startup, so it's fine that the database doesn't yet have the `metadata` table (though this is the last time it will be fine, since we are moving to multiple instances).

### Side note

There is no test that a process can hold the `Lock` for longer than the usual database timeout of 15s, which we certainly need to for long processing runs. But, it's easy to verify this is possible using IEx:

```elixir
AlertProcessor.Lock.acquire(fn :ok -> Process.sleep(60_000) end)
```


apps/alert_processor/lib/model/metadata.ex
```diff
@@ -0,0 +1,46 @@
...
+
+  @primary_key {:id, AlertProcessor.AtomType, autogenerate: false}
+
+  schema "metadata" do
+    field(:data, :map, default: %{}, null: false)
+    timestamps()
+  end
+
+  @spec get(key) :: value
+  def get(id) when is_atom(id) do
```
arkadyan, line 24: It seems like these 3 functions could use some quick tests?

apps/alert_processor/test/alert_processor/alert_parsing/alert_worker_test.exs
```diff
@@ -1,9 +1,120 @@
...
-  test "worker calls alert parser" do
-    {:noreply, _} = AlertWorker.handle_info({:work, 0, DateTime.utc_now()}, nil)
-    assert_received :processed_alerts
+  setup do
+    # Required due to Lock using a spawned process
+    Ecto.Adapters.SQL.Sandbox.mode(AlertProcessor.Repo, :auto)
+    Metadata.delete(AlertWorker)
+  end
+
+  describe "server" do
```
arkadyan, line 13: I love all these new tests.

arkadyan (commented):
This is great and so exciting to see!

I asked one clarifying question and pointed out a place where maybe you could add a few tests, but otherwise, this is looking really good.

apps/alert_processor/lib/lock.ex
```diff
@@ -0,0 +1,71 @@
...
+                end
+
+              %{rows: [[false]]} ->
+                send(parent, :not_acquired)
+            end
+          end,
+          timeout: :infinity
+        )
+
+        send(parent, :test_done)
```
digitalcora, line 46: oops, this was left over from an earlier test. Removing...

apps/alert_processor/lib/lock.ex
```diff
@@ -0,0 +1,71 @@
...
+
+        Repo.transaction(
+          fn ->
+            case Repo.query!("SELECT pg_try_advisory_xact_lock($1)", [@key]) do
+              %{rows: [[true]]} ->
+                send(parent, :acquired)
+
+                receive do
+                  :release -> nil
+                  {:EXIT, _, _} -> nil
```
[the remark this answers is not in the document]
digitalcora: > Does this handle any case where the process holding a lock dies or is killed?

In general yes, though how it happens depends on what you mean by "the process holding a lock".

1. If any code in _this_ process (the "locker") raises an exception somehow, `Repo.transaction` should guarantee the transaction is rolled back, which releases the advisory lock.
2. If this process dies or is killed, the `trap_exit` flag above means it will get an `:EXIT` message here, releasing it from the `receive` and closing the transaction, which again releases the advisory lock.
3. If the process calling `acquire` raises an exception inside the function it passed to `acquire`, the `try ... after` below will notify the "locker", releasing it from the `receive` and closing the transaction as in (2).
4. If the process calling `acquire` dies or is killed, the "locker" will also be terminated due to it having been started with `spawn_link`, and then (2) applies.

The test you linked is specifically testing (3). There's currently no explicit test for (4), though [this test](https://github.com/mbta/alerts_concierge/pull/1145/files#diff-fbfbcd8ddf827e97dff2425505302219eee6815e764ab368378d07e362c67a2fR26) actually tests it indirectly (the test process getting torn down before the `sleep` has finished creates this situation). I think adding a test for it would be good, anyway.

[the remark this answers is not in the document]
arkadyan: 👍 thanks, that's a helpful description.

apps/alert_processor/lib/model/metadata.ex
digitalcora: Yes, they could! Forgot all about it. 🙃

digitalcora:
@arkadyan Just so it's seen by more than one pair of eyes, I added a small config change to this branch after discovering an error that occurred on the dev-green deployment. The commit message should explain it!

arkadyan:
@digitalcora 👌

PR #1150, opened 2021-05-14
build: remove Codecov integration

digitalcora:
Once this project is migrated to GitHub Actions we can add the coverage reporter we've been using on other projects, but we need to remove this now since our Codecov subscription is expiring.

Coverage reports can still be generated locally with `MIX_ENV=test mix coveralls.html`, though unfortunately this doesn't provide an easy way to see how a branch will change coverage.


arkadyan (approved):
👍 

When you are ready, I think you should be able to [enable](https://github.com/mbta/alerts_concierge/actions/new) `Report Elixir coverage`—and `Elixir CI` which it depends on—and replace `excoveralls` with `lcov_ex` and the action should work for you. [Example here.](https://github.com/mbta/auto_pay_admin/pull/133)

PR #1152, opened 2021-06-03
test: fix failures due to removal of old stop IDs

digitalcora:


arkadyan (approved):
👍

apps/alert_processor/test/alert_processor/alert_parsing/alert_parser_test.exs
```diff
@@ -517,7 +517,7 @@ defmodule AlertProcessor.AlertParserTest do
 
         assert %{
                  departure_time: "2017-10-26T08:52:00-04:00",
-                 stop_id: "place-ER-0362",
+                 stop_id: "Newburyport",
```
digitalcora, line 520: I'm not sure why this actually worked with the "new" stop ID, since it's using a VCR cassette with data that includes the "old" stop ID — and equally unsure why it would stop working now. The fact that it broke right when the real-world data changed points to the possibility that the VCR setup is not working reliably 😬

arkadyan: Yes, based on things I've seen with this project and similar issues in dotcom, I strongly suspect VCR (at least the way we have it set up) isn't properly working.

digitalcora: The version of the library on this project is also (like most other deps) a couple years old, so I'm hoping future dependency updates will improve the situation.

PR #1153, opened 2021-06-03
test: fix rare failure in `Lock` test

digitalcora:
Noticed this failure for the first time in the CI run for #1152, so it's pretty rare, but possible.

`Process.exit` is not synchronous, and this test did not wait for the process to exit before continuing, so occasionally it would fail because the lock was not released by the time the test tried to acquire it.


arkadyan (approved):
👍

PR #1154, opened 2021-06-03
migrate: add email rejection field to users

digitalcora:
This introduces a field that will be used for handling email rejections — it has to be deployed separately, and before the main changes, so the field will be in place before any code tries to use it.


arkadyan (approved):
👍 

Is AWS's model is that they only tell us about a rejection status after going through any retries? So getting a rejection status is enough for us to know to stop sending to that email?

digitalcora:
> Is AWS's model is that they only tell us about a rejection status after going through any retries?

Correct: if we get notified about a transient bounce, it's already been retried several times. But also, in the upcoming implementation, we only disable email sending on a permanent bounce. Transient bounces can occur for reasons like inbox storage being full, so we probably want to keep trying.

PR #1155, opened 2021-06-03
Handle email bounces and complaints

digitalcora:
We had partially set up an endpoint intended to handle these messages, but it didn't do anything other than log them, didn't verify signatures, and was not subscribed to the SNS topic anyway.

This revamps the endpoint to correctly respond to complaints and permanent bounces by setting the user's communication mode to `none`, similar to how SMS opt-out works. This prevents sending any further notifications until the user next updates their account. We also set an `email_rejection_status` field, both to indicate the reason the mode is `none` (since there are now multiple reasons it could be), and to enable better UI for this feature later.

Currently when the mode is `none` it will appear to be "Email" on the account page. This was also an issue with SMS opt-out, so it can be addressed separately, and shipping this change sooner is better so we can start reducing our bounce/complaint rates.

### Other changes

I encountered an issue with `ExAws.SNS` where the typespec for the `verify_message` function (which we start using here) is actually wrong. The best solution I could find for this was to ignore the whole library, which required upgrading Dialyxir, so I did that as a separate commit. See: https://github.com/ex-aws/ex_aws_sns/pull/16#issuecomment-853963203


apps/concierge_site/lib/controllers/rejected_email_controller.ex
```diff
@@ -4,35 +4,84 @@ defmodule ConciergeSite.RejectedEmailController do
...
+         "bounceType" => "Permanent",
+         "bouncedRecipients" => [%{"emailAddress" => email}]
+       }) do
+    get_and_update_user(email, &User.set_email_rejection(&1, "bounce"))
   end
 
-  defp do_handle_rejected_email(conn, %{"notificationType" => "Complaint", "complaint" => params}) do
-    log_rejected_email(params["complainedRecipients"])
+  # Take no action on other bounce types, but assert the type is one of the expected ones
+  defp handle_bounce(%{"bounceType" => type}) when type in ~w(Transient Undetermined), do: nil
```
arkadyan, line 64: Do we want to consider logging these just for awareness or the ability to look them up later?

arkadyan (approved):
👍 This looks great. Noted one area where you might consider adding logging if it seems like a good idea to you, but otherwise :shipit:

apps/concierge_site/lib/controllers/rejected_email_controller.ex
digitalcora: We are logging all notifications on L41, so that will catch these as well.

arkadyan: Oops, yep missed that 👍

PR #1156, opened 2021-06-16
refactor: update Credo and fix some warnings

digitalcora:
Since this was a pretty big version jump, this resets the configuration to the new default and resolves some resulting warnings:

* Missing `@moduledoc` attributes
* Unnecessary `with` where a `case` would work instead
* Unnecessary `with` due to all statements being assignments
* Inefficiently checking `length(...) == 0` instead of `Enum.empty?`

Remaining warnings are disabled so `mix credo` will be clean for CI, which we will start doing as part of the move to GitHub Actions.


arkadyan (approved):
👍 That's a lot of tedious clean-up, thanks.

PR #1158, opened 2021-06-22
test: fix failures due to CR schedule changes

digitalcora:


apps/concierge_site/test/lib/schedule_test.exs
```diff
@@ -265,8 +260,6 @@ defmodule ConciergeSite.ScheduleTest do
                  ]
                }
              }
-
-    assert Enum.find_index(return_trip_result[{"cr", "CR-Worcester"}], & &1.weekend?) == 4
```
digitalcora, line 269: I removed these assertions outright because it wasn't clear what, if any, value they were providing (why would the specific index of a trip result be important?), and it seems like there's a very low risk that something would regress because this assertion wasn't here.

arkadyan: Make sense. My only guess is that they might have been attempting to provide coverage around the `weekend?` function, but this isn't the way to do it.

arkadyan (approved):
👍

PR #1159, opened 2021-06-23
ci: add no-op deploy workflow

digitalcora:
`workflow_dispatch` workflows need to exist in some form on the repo's main branch to be available, so this adds a "deploy" workflow that does nothing, to enable testing of the real deploy workflow on a branch.


arkadyan (approved):
Clever!

PR #1160, opened 2021-06-23
ci: enable deployment with Actions

digitalcora:
See [successful dev-green deploy](https://github.com/mbta/alerts_concierge/runs/2898382424?check_suite_focus=true) from this branch!

Some notable "innovations" I'd like to backport or at least socialize:

* This allows manually deploying to any environment, and auto-deploying the main branch to `dev`, using a common workflow. The UI isn't great in that you have to [type in the name of the environment](https://user-images.githubusercontent.com/394835/123157928-2857d000-d439-11eb-8df5-8ce5b2ed76d0.png) instead of selecting it from a list — but if you typo, nothing terrible happens (it does pollute the list of "Environments" in the repo settings).

* The auto-deploy only happens after CI finishes, and only if it passed... at least, it _should_ work this way, but it doesn't appear there's any way to test it until the new config is committed to the main branch. So, we'll see. 🤞

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,8 +1,10 @@
 name: Deploy
 
 on:
-  push:
-    branches: master
+  workflow_run:
+    workflows: ["CI"]
+    types: [completed]
+    branches: [master]
   workflow_dispatch:
     inputs:
       env:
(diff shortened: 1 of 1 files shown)


arkadyan (approved):
Just to point out one tradeoff with a single build is that you don't have the ability (I assume?) to configure dev separately, for example, to automatically deploy after any PR merge as many projects do. You probably already realize that and are taking it into account. This looks good to me.

digitalcora:
> you don't have the ability (I assume?) to configure dev separately, for example, to automatically deploy after any PR merge

The workflow does in fact have this specific ability, the "auto-deploy" I mentioned. At least, we'll find out if it does shortly 😄

arkadyan:
@digitalcora Sorry! Yes, you very clearly mentioned that specifically and apparently it (metaphorically) went in one ear and right out the other.

digitalcora:
Looks like the auto-deploy-to-dev [worked](https://github.com/mbta/alerts_concierge/actions/runs/965729422) 🚀

PR #1162, opened 2021-06-28
ci: add cancel-in-progress concurrency option

digitalcora:
`github.ref` here is a branch or tag name (not a commit hash), so this means if we push new commits to a branch that already has an ongoing CI job, it will be cancelled in favor of the new one.


arkadyan (approved):
Cool!

PR #1163, opened 2021-07-15
Improve UI and explanations when alerts are disabled

digitalcora:
Currently a `none` communication mode is displayed the same as `email`, making it appear email delivery is enabled when in fact it isn't. There's no other indication in the UI that alerts are disabled, or explanation as to why, except when the reason is SMS opt-out and the user is still inside the opt-out freeze period.

With these changes, if a user's communication mode is `none`:

* their Settings page correctly has neither of the delivery options selected

* a flash appears on their Settings page explaining how this happened and what they can do to resolve it (for all circumstances, not just when inside the SMS freeze period)

* a flash appears on their Subscriptions page, which is the "home page" when signed in, explaining that alerts are disabled and to access the Settings page to resolve this

### DIFF
--- a/apps/concierge_site/lib/controllers/trip_controller.ex
+++ b/apps/concierge_site/lib/controllers/trip_controller.ex
@@ -17,9 +17,20 @@ defmodule ConciergeSite.TripController do
   @valid_facility_types ~w(bike_storage elevator escalator parking_area)
 
   def index(conn, _params, user, _claims) do
-    trips = Trip.get_trips_by_user(user.id)
-    render(conn, "index.html", trips: trips, user_id: user.id)
+    conn
+    |> communication_mode_flash(user)
+    |> render("index.html", trips: Trip.get_trips_by_user(user.id), user_id: user.id)
   end
+
+  defp communication_mode_flash(conn, %User{communication_mode: "none"}) do
+    put_flash(
+      conn,
+      :warning,
+      "Alerts are disabled for your account. To resume alerts, check your account Settings."
+    )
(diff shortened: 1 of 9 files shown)


apps/concierge_site/lib/controllers/trip_controller.ex
```diff
@@ -17,10 +17,21 @@ defmodule ConciergeSite.TripController do
...
+    conn
+    |> communication_mode_flash(user)
+    |> render("index.html", trips: Trip.get_trips_by_user(user.id), user_id: user.id)
+  end
+
+  defp communication_mode_flash(conn, %User{communication_mode: "none"}) do
+    put_flash(
+      conn,
+      :warning,
+      "Alerts are disabled for your account. To resume alerts, check your account Settings."
```
arkadyan, line 29: Would it be helpful if this included a link to your account settings?

arkadyan (approved):
Looks great to me 👍

apps/concierge_site/lib/controllers/trip_controller.ex
digitalcora: Currently flash content can only be plain text (it is HTML-escaped on render), so before making that possible we'd have to go through and ensure all flash content is HTML-safe. I might look into that anyway since I am also waiting on a content review.

arkadyan: Ah right, I forgot about that. Well I'll leave it up to you, but it doesn't seem that important.

PR #1164, opened 2021-08-02
Fix multiple instance issues with SMS opt-out

digitalcora:
As an apparent side effect of migrating the app to multiple instances, the worker that fetched the SMS opt-out list stopped working. This was due to a combination of factors:

* the rate limit for this endpoint is very low (10 requests/sec)
* our opt-out list is very large (~12,000 entries), requiring over a hundred "paginated" requests to fully retrieve
* AWS seems to start up new instances of an app at almost the same time

The result was that while one instance would successfully fetch the list, if two instances were running at once, they would both try to fetch at the same time, and both fail due to the rate limit being used up twice as fast.

Though it's not exactly the same problem since AWS rate limits are account-wide, we can largely solve this by reusing the exclusive-lock approach from the original multiple-instance work.


arkadyan (approved):
👍 nice

PR #1166, opened 2021-08-19
Delete Mailchimp data when account is deleted

digitalcora:
This was a request from the team drafting new MBTA privacy policies. Currently when a user self-service deletes their account, we permanently delete all data connected to them _except_ their membership on the weekly digest in Mailchimp, where we only mark them as unsubscribed. It will simplify the privacy policy (and also better reflect the intent of the action) if deleting your account truly deletes everything, including Mailchimp data.

This change is deployed to dev-green, where I tested that existing functionality still works and that the new delete request goes through.

### DIFF
--- a/apps/concierge_site/lib/mailchimp.ex
+++ b/apps/concierge_site/lib/mailchimp.ex
@@ -1,96 +1,71 @@
 defmodule ConciergeSite.Mailchimp do
   @moduledoc """
-  functions for sending unsubscribes and subscribes
+  Handles updating digest subscribership in Mailchimp and receiving updates from the service.
   """
 
   require Logger
   alias AlertProcessor.Helpers.ConfigHelper
   alias AlertProcessor.Model.User
   alias AlertProcessor.Repo
-  alias HTTPoison
+  alias Ecto.Changeset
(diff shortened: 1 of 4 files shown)


apps/concierge_site/lib/mailchimp.ex
```diff
@@ -1,96 +1,84 @@
...
   require Logger
   alias AlertProcessor.Helpers.ConfigHelper
   alias AlertProcessor.Model.User
   alias AlertProcessor.Repo
-  alias HTTPoison
+  alias Ecto.Changeset
+
+  @doc "Update a user's subscriber status in Mailchimp to reflect their `digest_opt_in` value."
+  @spec update_member(User.t()) :: :ok | :error
+  def update_member(%{id: id, email: email, digest_opt_in: digest_opt_in}) do
```
arkadyan, line 14: I wonder if the method was called something like `upsert_member` if that might help it make sense when we use it for both adding and updating.

arkadyan (approved):
👍 

I left a comment about a function name, but I don't have strong feelings about it, take it or leave it.

PR #1170, opened 2021-08-23
refactor: remove unused "header note" feature

digitalcora:
This was previously used to display a warning about the govDelivery version of T-Alerts being turned off. It is not referenced anywhere in the code since September 2018.


arkadyan (approved):
👍 🧹

PR #1172, opened 2021-09-09
build: upgrade to latest Node LTS release

digitalcora:
Thankfully it seems like there are no incompatibilities here other than the Yarn stuff I saw before (and addressed in #1167), so this wasn't much more than increasing a version number. The `package-lock` changes are due to the new version of NPM generating the file differently, but should not have changed any actual installed packages.


arkadyan (approved):
👍

PR #1174, opened 2021-09-14
feat: remove unused "courtesy email" feature

digitalcora:
The currently configured address for these emails has been bouncing for at least the past few months. To our knowledge, no one has missed them, and the bounces are a source of error noise.


arkadyan (approved):
👍 

Is `COURTESY_EMAIL_ADDRESS` a variable defined on the servers? If so I'm assuming it can be deleted after this is deployed. Do you know if you have the permissions to do that or will we need a devops ticket or something?

digitalcora:
I should be able to do that, and plan to once this is deployed to prod.

PR #1175, opened 2021-09-14
feat: improve password character restrictions

digitalcora:
Previously we placed several odd restrictions on passwords that were also not communicated well (most resulted in a browser popup saying something like "Please match the requested format"). It's unclear what security benefit these had, and they presented a frustrating obstacle to creating an account or resetting a password even for people using password-management best practices, due to forbidding many characters commonly found in randomly-generated passwords.

This change:

* removes the browser-side input validation entirely, since it is redundant with the server-side validation and it's not possible to customize the error message without adding some JavaScript

* simplifies the server-side validation to only require at least 8 characters (up from 6, though this is still rather short by modern standards), and at least one number or symbol


arkadyan (approved):
Great!

PR #1176, opened 2021-09-22
Update Webpack and frontend build tools

digitalcora:
* Webpack is upgraded from v4 to v5, and most other devDependencies are upgraded to their latest versions. This resolves a number of security warnings.

* `node-sass` is deprecated in favor of `sass`.

* Using `/` for division in Sass is deprecated in favor of `math.div`, though in the one case we used this in our own code, it's more easily replaced with `* 0.5`. We enable `quietDeps` since this deprecation otherwise causes a huge number of warnings from Bootstrap.

* `optimize-css-assets-webpack-plugin` is not available for Webpack 5, and is replaced by `css-minimizer-webpack-plugin`.

* `babel-polyfill` is deprecated in favor of explicitly installing `core-js` and specifying the chosen version in `.babelrc`.

* Several packages were identified as unused using `depcheck`, and removed.

* An `npm build` script is added to allow building the assets in development mode without starting the watcher.

This is currently deployed to `alerts-dev-green`, and the frontend there seems to still look and work as it should.


arkadyan (approved):
👍

PR #1177, opened 2021-09-22
build: Dockerfile improvements

digitalcora:
* Use pre-built `hexpm/elixir` image instead of starting from an Erlang image and manually installing Elixir
* Use `node` image to build assets instead of installing Node in the Elixir image (allows specifying exact version)
* Upgrade to current stable Debian release ("Bullseye")
* Use `slim` Debian images to omit unneeded packages where possible
* Use `dumb-init` as entrypoint to ensure signals are handled correctly

Overall, these improvements reduce our deploy time by about 2 minutes, and better align with how we're currently doing Docker builds on other projects like Dotcom and Arrow.


arkadyan (approved):
👍

PR #1178, opened 2021-10-12
deps: update and clean up frontend packages

digitalcora:
* `elem-dataset` is replaced with native `dataset` since all browsers now support it, even IE11
* `whatwg-fetch` polyfill is removed as all browsers other than IE11 support `fetch` natively, and this function is only used in the admin interface
* All package updates otherwise had no breaking changes for us

Packages that remain outdated:

* `react`, `react-dom`, and `typescript`: these are used only in the admin interface, and we're not yet sure whether we will keep this
* `bootstrap`: this is already at the latest release of Bootstrap 4, and migrating to Bootstrap 5 is pretty far out of scope for this task

### DIFF
--- a/apps/concierge_site/assets/js/admin/api/index.ts
+++ b/apps/concierge_site/assets/js/admin/api/index.ts
@@ -1,5 +1,3 @@
-import 'whatwg-fetch';
-
 interface UserResponse {
   id: string;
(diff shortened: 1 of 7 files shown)


arkadyan (approved):
👍

PR #1179, opened 2021-10-14
Fix missing favicon

digitalcora:
This occurred because the `<link rel="icon">` in the HTML used the "digested" filename in production, but following this link would give a 404, since the site was only configured to serve the un-digested `favicon.ico`. Depending on whether the browser "preferred" the `<link>` or the file at `/favicon.ico`, it sometimes wouldn't show the icon.

Using `only_matching` instead of `only` in `Plug.Static` allows serving digested files at the root, since strings specified here only need to be a prefix of the first path component instead of matching it exactly.

### DIFF
--- a/apps/concierge_site/lib/endpoint.ex
+++ b/apps/concierge_site/lib/endpoint.ex
@@ -17,7 +17,7 @@ defmodule ConciergeSite.Endpoint do
     at: "/",
     from: :concierge_site,
     gzip: true,
-    only: ~w(css fonts images js favicon.ico robots.txt analytics.txt)
+    only: ~w(css fonts images js favicon.ico robots.txt)
   )
 
   # Code reloading can be explicitly enabled under the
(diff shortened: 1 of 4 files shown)


apps/concierge_site/lib/endpoint.ex
```diff
@@ -17,7 +17,7 @@ defmodule ConciergeSite.Endpoint do
     at: "/",
     from: :concierge_site,
     gzip: true,
-    only: ~w(css fonts images js favicon.ico robots.txt)
+    only_matching: ~w(css fonts images js favicon robots)
```
arkadyan, line 20: TIL

arkadyan (approved):
👍

apps/concierge_site/lib/endpoint.ex
digitalcora: Same! The docs for this are surprisingly vague. I had to dig into the source to even understand that what it was "matching" against (in both cases) is just the first path component.

PR #1182, opened 2021-11-17
ci: install Rebar and Hex on ASDF cache miss

digitalcora:
Since these tools are "global" and part of the Elixir install, they should be reinstalled if we have a new language version, not if we have a new version of the project dependencies.

See: https://github.com/mbta/.github/pull/11

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,12 @@ jobs:
         if: steps.asdf-cache.outputs.cache-hit != 'true'
       - uses: mbta/actions/reshim-asdf@v1
 
+      - name: Elixir tools
+        if: steps.asdf-cache.outputs.cache-hit != 'true'
+        run: |
+          mix local.rebar --force
+          mix local.hex --force
+
       - uses: actions/cache@v2
         id: ex-deps-cache
         with:
@@ -46,8 +52,6 @@ jobs:
       - name: Elixir dependencies
         if: steps.ex-deps-cache.outputs.cache-hit != 'true'
         run: |
-          mix local.rebar --force
-          mix local.hex --force
           mix deps.get
           mix deps.compile
 


.github/workflows/ci.yml
```diff
@@ -35,6 +35,13 @@ jobs:
       - uses: asdf-vm/actions/install@v1
         if: steps.asdf-cache.outputs.cache-hit != 'true'
       - uses: mbta/actions/reshim-asdf@v1
+        if: steps.asdf-cache.outputs.cache-hit == 'true'
```
paulswartz, line 38: You always need to do this, as it also sets up the `PATH` environment variable to point at the ASDF versions of Elixir/Erlang.

digitalcora: Oh, I didn't expect we'd have to, since I assumed `asdf-vm/actions/install` would fully set things up. Maybe we should name our action something other than `reshim-asdf` since it does more than just reshim?

.github/workflows/ci.yml
```diff
@@ -46,8 +52,6 @@ jobs:
       - name: Elixir dependencies
         if: steps.ex-deps-cache.outputs.cache-hit != 'true'
         run: |
-          mix local.rebar --force
-          mix local.hex --force
           mix deps.get
```
arkadyan, line 55: Just to double check, in https://github.com/mbta/.github/pull/11 it adds `if: if: steps.deps-cache.outputs.cache-hit != 'true'` to `mix deps.get`—we're good with skipping that bit?

arkadyan (approved):
Added one question based on comparing with the source PR, but I'm good with this change as is if you think that's best.

.github/workflows/ci.yml
digitalcora: We already have that here, see L53. (though thanks for discovering that `if: if:` typo in the other PR 😄)

PR #1184, opened 2021-12-08
feat: new administration portal

digitalcora:
This is a complete replacement of the previous admin portal, with the new one being built using server-rendered views similar to the rest of the app, where the old one used React, TypeScript, and an internal JSON API. Since these dependencies were only used by the admin portal, this allows us to delete them entirely.

The "Administration" page now offers a menu of functions:

* "Database queries" allows running predefined SQL queries and viewing the results in a table or exporting them to CSV. This replaces the queries committed to `report_queries` which had to be run manually by someone with direct database access, and also the `Metrics` which were logged periodically to allow inspection in Splunk.

* "Manage admin access" allows granting or revoking admin privileges, and viewing the email addresses of all current admins. Previously there was no way to change user roles without direct database access.

* The previous function of the admin portal, searching for users and deleting them, is not retained. Considering users can delete their own accounts, it hasn't really been needed in recent history, and it exposes personal information.

### Screenshots

Home | Admins | Queries | Query
--- | --- | --- | ---
Screen Shot 2021-12-08 at 3 35 42 PM | Screen Shot 2021-12-08 at 3 35 55 PM | Screen Shot 2021-12-08 at 3 36 40 PM | Screen Shot 2021-12-08 at 3 37 02 PM


arkadyan (approved):
Looks great! This is a whole lot of useful features presented in a much more handy way.

PR #1185, opened 2021-12-09
refactor: Remove an unused test helper module

arkadyan:
This is a pure drive-by: I happened to stumble on this file and noticed it wasn't being used.


digitalcora:
A surprise, to be sure, but a welcome one 😄

PR #1186, opened 2021-12-13
feat(admin): report for subscription count by stop

digitalcora:
Temporarily deployed to production to ensure this is reasonably optimized — the query completes in less than a second.


arkadyan (approved):
👍

PR #1188, opened 2022-01-06
feat(admin): add column to users report

digitalcora:
This adds a column `subscribed_users` which is the count of users who have any subscriptions, regardless of paused status.


arkadyan (approved):
👍

PR #1189, opened 2022-01-06
build: remove "app version" generation

digitalcora:
It's unclear whether this still worked, and in any case we don't use it for anything.

### DIFF
--- a/apps/alert_processor/mix.exs
+++ b/apps/alert_processor/mix.exs
@@ -4,7 +4,7 @@ defmodule AlertProcessor.Mixfile do
   def project do
     [
       app: :alert_processor,
-      version: "0.0.59",
+      version: "0.1.0",
       build_path: "../../_build",
       config_path: "../../config/config.exs",
       deps_path: "../../deps",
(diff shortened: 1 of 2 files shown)


apps/alert_processor/mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule AlertProcessor.Mixfile do
   def project do
     [
       app: :alert_processor,
-      version: "0.0.59",
+      version: "0.0.1",
```
arkadyan, line 7: suggestion: Would it make sense to bump these to `1.0.1` or something so that they are increasing?

arkadyan (approved):
👍

I added a comment because lowering the version number felt sad to me for some reason, but I don't really care at all.

apps/alert_processor/mix.exs
digitalcora: Ah, I should set it to `0.1.0`, since that's the default version when you generate a new Elixir project (and so it's what most of our other apps have).

PR #1190, opened 2022-01-07
refactor: remove dependency version overrides

digitalcora:
None of these overrides were actually required to resolve dependencies, except `tzdata` which was overridden to a higher version than supported by `calendar` and `fast_local_datetime`. These packages have versions that support newer `tzdata` with no backwards compatibility issues for us, so we can update them.


arkadyan (approved):
👍

PR #1191, opened 2022-01-07
refactor: clean up `extra_applications`

digitalcora:
Almost none of these were necessary for our current versions of Elixir and dependencies.


arkadyan (approved):
👍

PR #1192, opened 2022-01-13
refactor: update Guardian from 0.x to 1.x

digitalcora:
This was one of the trickier backwards-incompatible upgrades blocking our migration from Ecto 2 to 3. There is unfortunately no version of Guardian that supports both 2 and 3, so this gets us as close as we can without upgrading Ecto.

See also the Guardian 1.0 upgrade guide: https://github.com/ueberauth/guardian/blob/3735af1/upgrade_guides/0.14.to.1.0.md

Notable changes for us:

* When we originally set up `guardian_db`, we never added the token sweeper worker that deletes expired tokens from the database. This is now added.

* Guardian 1.0 removes `Guardian.Phoenix.Controller`, though for some reason this is not documented in the upgrade guide. We actually didn't need to use this, since we already have a plug that assigns the current user into the conn.

* Previously, when creating or updating a subscription, if the current auth token claims had an `imp` key, the user with that ID would be treated as the "originator" of the update for PaperTrail purposes. This is removed, as it had no apparent purpose and nothing else in the codebase or in Guardian itself references this key. It may have been a remnant of an "impersonation" feature.

* Instead of manually assigning permissions to a user in every place where a "sign-in" happens, we use the new `build_claims` callback.

* The `TokenLogin` plug is deleted, as the only place it was used was the tests for `TokenRefresh`. Those tests are rewritten to use the same Guardian functions the real app does.


arkadyan:
_Initial tiny suggestion after reading only the first paragraph of your description:_ I was curious about what version of Guardian we'd need to upgrade to when upgrading to Ecto 3 and it occurred to me that maybe you'd find it helpful to jot that down in https://app.asana.com/0/1167196461144361/1198926652145357 or somewhere while it's front of mind.

digitalcora:
@arkadyan The next step after this would be to upgrade to the latest version of Guardian — there won't be another intermediate upgrade (unless Guardian comes out with a 3.x major release before we get to it).

arkadyan (approved):
👍 Nice. Thank you for digging through all these details.

PR #1193, opened 2022-01-14
Add Sobelow and fix identified vulnerabilities

digitalcora:
See commit messages for details.


arkadyan (approved):
👍 Looks great.

This is a good reminder that I should add Sobelow to the list of recommended tools to consider using in new apps.

PR #1195, opened 2022-01-19
fix: 500 error when auth token expired

digitalcora:
Sentry report: https://sentry.io/organizations/mbtace/issues/2930926732/?environment=prod&project=5583106

Guardian 0.14.x had a default set of error handlers it would fall back to if you didn't handle a particular type of error. This is no longer the case in Guardian 1.x due to the error handler API changing to use different clauses of a single callback, instead of multiple callbacks. This means we need to include a fallback clause ourselves.


arkadyan (approved):
👍

PR #1197, opened 2022-01-27
Expand rate-limiting and fix remote IP forwarding

digitalcora:
The main change here is to expand rate limiting from just the sign-in form to any action that submits data (is not a GET request). This primarily prevents scripts from abusing the sign-_up_ form to tank our email sender reputation, because we have to send at least one email that bounces before we get a bounce notification letting us know we should disable emails. We also plan to guard against this by adding reCAPTCHA to the sign-up form, though.

Since the new solution is global instead of per-action, for now we have no use for `hammer_plug` anymore, since it does not provide conveniences for global rate limiting based on the method of the request.

This also fixes an issue where the "remote IP" used for rate limiting (and incidentally in Sentry reports) would always be the internal IP of the AWS load balancer, which could result in over-limiting.


apps/concierge_site/lib/plugs/rate_limit.ex
```diff
@@ -0,0 +1,43 @@
...
+  import Plug.Conn
+  @behaviour Plug
+
+  @limit_anonymous 5
+  @limit_signed_in 15
+  @one_minute 60_000
+
+  @doc """
+  Initialize the plug. If the option `enable?: false` is given, no actual rate limiting will be
+  performed except for requests with a `rate_limit?: true` assign in the conn.
```
arkadyan, line 13: **suggestion (non-blocking):** Personally I would find it helpful to have a super quick hint on why/when the `rate_limit?: true` condition might occur. Is it just for testing purposes?

arkadyan (approved):
Add a tiny suggestion if you want to take it, but otherwise 👍 — this is slick.

PR #1200, opened 2022-01-31
feat(admin): filter inactive users out of reports

digitalcora:
Following a vulnerability scan that created a massive number of fake accounts, complete with fake subscriptions, the numbers in our reports were a bit skewed. We can compensate for this by ignoring accounts where alerts are disabled, which happens when either their email address bounces (which we saw with these accounts, since they made up non-existent addresses) or they opt out via SMS. In cases where this occurs "naturally", it should still be okay to not include those accounts in the reports, since we are primarily interested in subscription stats for users where those subscriptions are actually generating notifications.


arkadyan (approved):
👍

PR #1201, opened 2022-02-01
fix: add fallback clause for account create

digitalcora:
When automated spam-bots try to create an account without even loading the JavaScript for the reCAPTCHA, the required parameter is missing and we get a crash report because no clause matches in the controller. This can also happen legitimately if the user's browser is old enough to not be supported by reCAPTCHA or is blocking JavaScript, so we can provide a nice error message.


arkadyan (approved):
👍

PR #1202, opened 2022-02-02
deps: upgrade to Phoenix 1.4 and Ecto 3

digitalcora:
> 🟢 This branch is deployed to dev-green, and I've done a manual smoke test of creating a subscription, creating a test alert that matched the subscription, and receiving the alert over both email and SMS.

This is mainly the Ecto 3 upgrade, but since few libraries that depend on Ecto support 2 and 3 within a single version, several other upgrades are required, including Phoenix.

See also the [Phoenix 1.3 to 1.4 upgrade guide][1].

Oddly there is no official Ecto 3 upgrade guide. [This blog post][2] describes a subset of the breaking changes, and they are all enumerated in the [CHANGELOG][3].

[1]: https://gist.github.com/chrismccord/bb1f8b136f5a9e4abc0bfc07b832257e
[2]: https://dashbit.co/blog/a-sneak-peek-at-ecto-3-0-breaking-changes
[3]: https://github.com/elixir-ecto/ecto/blob/master/CHANGELOG.md#backwards-incompatible-changes-1

Notable changes for us:

* Ecto 3 introduces a formal separation between datetimes with and without fractional seconds, and strictly validates these when storing values (though when loading they are just truncated, if applicable). This requires that we make a decision one way or the other about which type our "datetime" columns actually are. In practice there is no field where fractional seconds are important, and this also turns out to be the easier of the two choices in terms of changes required to existing code. The main gotcha is that `DateTime.utc_now()` returns fractional seconds, so we can't assign it directly to a field declared as `utc_datetime`.

* Related to the above, since we never declared a type for `timestamps` columns, they were defaulting to `naive_datetime`. For consistency with other datetime fields, we change these to `utc_datetime`. This works without any data migrations because the "local" time zone for both RDS databases and ECS containers defaults to UTC (so actually it doesn't matter which type we declare it to be, but since we know it's UTC, better to be specific about that).

* Though we had `tzdata` installed for some time, we never set Elixir's `:time_zone_database` config value to use it. This simplifies some test code.

* Phoenix 1.4 introduces a `mix format` integration. This means the CI action for formatting needs to load the dependency cache.

* The `Ecto.Type` behaviour has some new required callbacks, which we need to implement for our custom `AtomType`.

* Functions passed to `Ecto.Multi.run/3` are now expected to accept two arguments (repo, changes) instead of one (changes).

* "Raw" datetimes returned from `Repo.query` are now `NaiveDateTime` instead of tuples. This simplifies some logic in `QueriesView`.

* The `errors_on` test helper (part of the Phoenix-generated `DataCase`) is updated to its Phoenix 1.4 version to account for a new behavior of `validate_inclusion`.

* For unclear reasons, the IDs of hour and minute dropdowns generated using `time_input/3` were no longer what we expected, and we now have to manually specify them. It's possible we were accidentally relying on a now-fixed bug in `phoenix_html`, as this seems like something that should never have worked.


arkadyan (approved):
👍 Thanks for the comprehensive write-up.

PR #1204, opened 2022-02-15
fix: Change api-dev url

uchukwu:
Update hardcode api-dev url where applicable


paulswartz (commented):
This looks fine, but since https://api-dev.mbtace.com/ doesn't exist the tests are failing. You should be able to re-run the tests when it does.

losvedir:
https://api-dev.mbtace.com/ exists now, but the tests are still failing. @uchukwu - might be a little bit more to do on this PR.

PR #1205, opened 2022-02-23
fix: JS error in subscription creation

digitalcora:
Couple of other small commits which have their own explanations, but the original point of this was an investigation of a JS error I noticed. It turned out to be harmless, but is also fixed now, since doing so was trivial.


arkadyan (approved):
👍 nice catch

PR #1206, opened 2022-02-24
docs: explain `Lock` approach in light of Ecto 3

digitalcora:
This was originally implemented this way simply because `Repo.checkout` didn't exist in Ecto 2, and the idea was that we could simplify things by using a session-level lock after upgrading to Ecto 3. However, after further investigation, this is actually still the best way to do it. The comment now explains the rationale.

This also removes a `key >= 0` guard clause that was unnecessary, since lock keys can be any integer, including negative integers.


arkadyan (approved):
👍

PR #1207, opened 2022-02-25
deps: organize and update (almost) all dependencies

digitalcora:
Having previously gotten several tricky cases out of the way individually, The Great Update is upon us. 🌔 Updates that required no code changes are consolidated into a single commit, otherwise they were given their own commit. This also means they can be reviewed separately.

Dependencies that remain outdated after this are:

* `paper_trail` — could be updated slightly, but later versions require Elixir 1.11.
* `phoenix*` — will be updated separately as part of the Phoenix upgrade process.
* `poison` — already using the newest version allowed by our version of `postgrex`.
* `postgrex` — newer versions require Elixir 1.11.

So there will be one more round of dependency updates after we update to the latest version of Elixir, but it might be small enough that we can just fold it into that task.


apps/alert_processor/mix.exs
```diff
@@ -36,26 +36,24 @@ defmodule AlertProcessor.Mixfile do
...
       {:fast_local_datetime, "~> 1.0.0"},
       {:gettext, "~> 0.11"},
       {:httpoison, "~> 1.8.0"},
       {:paper_trail, "0.8.3"},
       {:poison, "~> 2.0"},
-      {:poolboy, ">= 0.0.0"},
       {:postgrex, "~> 0.15.0"},
       {:sentry, "~> 8.0"},
       {:sweet_xml, "~> 0.6"},
-      {:eflame, "~> 1.0", only: [:dev]},
```
arkadyan, line 59: 🔥

arkadyan (approved):
🌔

PR #1208, opened 2022-03-02
deps: upgrade to latest version of Phoenix

digitalcora:
With bonus commit for migrating away from `Supervisor.Spec`, since the Phoenix 1.5 upgrade instructions used a new-style child spec for its change to `PubSub`.


arkadyan (approved):
👍

PR #1209, opened 2022-03-03
deps: update to Erlang 24 and Elixir 1.13

digitalcora:
Notable changes:

* Switch to the new `Config` as `use Mix.Config` is deprecated
* Switch to the new `runtime.exs` config system
* Clean up some config code and remove outdated generated comments
* Add `poolboy` to deps, as we use it directly but didn't depend on it (now called out by the compiler)
* `mix format` now doesn't add line breaks inside interpolations to make long strings fit in the line length limit


arkadyan (approved):
👍

PR #1210, opened 2022-03-10
test: fix intermittent `SendingQueue` failures

digitalcora:
The queue is a global resource, so tests that use it cannot be `async`. We also need to make sure we `reset` it before such tests.


arkadyan (approved):
👍

PR #1211, opened 2022-03-10
fix: reversed route icon order in stop selectors

digitalcora:
Changes to attribute ordering due to the Phoenix upgrade in fe6c785c introduced a bug where route icons in stop selectors would display in the opposite of the intended order (see below). Since the reversed ordering is a consequence of using `float: right`, we can un-reverse it in the same place we apply this style.

Screen Shot 2022-03-10 at 11 55 02 AM


arkadyan (approved):
👍

PR #1214, opened 2022-04-05
Landing page adjustments

zovt:


digitalcora (approved):
Looks good, sorry it took so long for me to get to this 😅 Couple things to note:

1. I think these changes are small and related enough that they could be one commit, but in particular I think it's important that the copy change and the corresponding test change be in the same commit. It looks like you typically use squash-merge on `dotcom` so maybe you were already planning to do this.

2. Be sure to format the final commit messages according to [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) (if it was all one commit then perhaps just `feat: landing page adjustments`). I normally don't include the default `(#1214)` suffix that squash-merge appends to the message, since [GitHub already tracks the PR a commit came from](https://user-images.githubusercontent.com/394835/162465437-bff93b69-81c9-4e48-b16c-425eeaaeb017.png), but I don't object to it either.

3. Looking at the image change, I noticed at the upper end of the `sm` breakpoint (which you'd see on, for example, an iPad) the image seems [**L A R G E**](https://user-images.githubusercontent.com/394835/162464201-349a226e-03f3-455b-8cb7-2db372031947.png) compared to the other elements on the page. I wonder if there's an easy way to address this; maybe we could even simplify some of the styles currently used for `.landing-page__example`, which assumed the image would only ever display at higher breakpoints. I'll leave it up to you whether that's worth the effort, since Steve didn't mention it as an acceptance criterion.

digitalcora:
> It looks like squash-and-merge is the default here too

I actually almost always use rebase-merge, in combination with interactive-rebasing my commits on the branch to clean them up before merging. Usually the outcome is identical because the changes only warrant a single atomic commit anyway, but sometimes there's more to it (see e.g. #1203). I still have squash-merge enabled on the repo because sometimes it's more convenient in the single-commit case. ⚡

digitalcora (approved):
Looks good! ✨

PR #1215, opened 2022-04-08
Upgrade to API `2021-01-09`

zovt:
This PR updates our code to work with the latest version (`2021-01-09`) of the API. There were 2 major breaking changes I attempted to accommodate:

1. `shapes` no longer includes `stops` as a valid relationship, which we were relying on to deal with some Red Line branching logic
2. `departure_time` on schedules can now be `nil`

With regards to 1., I've removed the `route_shapes` function from `api_client`, and replaced it with `route_patterns_with_stops`, which fetches the pattern information along with the representative trip and stop data. This required adjusting the logic in `fetch_route_branches`. Since the data we get in `included` is a list with `trip` and `stop` data intermixed, we do some light massaging to separate the data by type and create some maps from ID -> data. We then filter the patterns by `typicality == 1` (which I believe should have a similar effect to what the previous `priority` filtering was doing). Then, for each of the patterns, we get the representative trip stop IDs for the pattern, using the maps we previous created (NOTE: I actually used the `parent_station` ID here, since the branch creation code expects IDs like `place-xxxx`. This feels pretty suspect to me, let me know if there's something better to do here). Ultimately this results in the red line branches getting created with the same data as before.

As for 2., there were a couple of places where we were sorting schedules based on `departure_time`. I've adjusted the sorter logic there such that any `nil` `departure_times` get sorted towards the end of the list.

I've kicked the tires here a little bit and haven't noticed anything breaking, but I definitely may have missed something here. Let me know if you run into anything major / there are better approaches to how I'm doing things!


digitalcora:
To clarify, we use shapes/patterns _only_ for getting the Red Line branches, and nowhere else? I suppose that does sort of make sense but it was a bit surprising to me.

> We then filter the patterns by `typicality == 1` (which I believe should have a similar effect to what the previous `priority` filtering was doing).

It's not exactly the same, but within this narrow use case I think it does hold true, and if it ever changed it would break a lot of other things anyway. (what we _really_ want is [canonical patterns](https://github.com/mbta/technology-docs/pull/6), but that's not implemented yet)

> (NOTE: I actually used the `parent_station` ID here, since the branch creation code expects IDs like `place-xxxx`. This feels pretty suspect to me, let me know if there's something better to do here)

No, that's correct. The app uses parent stations for pretty much all stop-related logic, converting child stops to parent stations eagerly. This is because we store stop IDs in our own database as part of subscriptions, and the IDs of parent stations tend to be more stable over time than child stops. Of course, it's not like parent station IDs never change, and we still have problems there, though this is at least [documented](https://app.asana.com/0/1167196461144361/1201516552243763).

> I've kicked the tires here a little bit and haven't noticed anything breaking, but I definitely may have missed something here. Let me know if you run into anything major / there are better approaches to how I'm doing things!

Two things I can think of:

1. We do have some `/schedules` API responses committed to the repo as test fixtures under `apps/alert_processor/test/alert_processor/fixture/custom_cassettes` (these are used by [ExVCR](https://github.com/parroty/exvcr)), so those might be using the old departure time rules. However, looking at a few of them, it seems like these might not be used anywhere in the current test suite. With maybe some quick automation we could determine this for sure (in order to use a cassette its name must be explicitly written in a `use_cassette` call). If it turns out we can delete them, we could throw that in as a separate commit within this PR? Or a separate PR if you prefer.

2. It'd be a good verification (and a good exercise for you!) to deploy this branch to dev-green and do an end-to-end smoke test, assuming you now have write access to the Alerts-UI test environment.

zovt:
> To clarify, we use shapes/patterns only for getting the Red Line branches, and nowhere else? I suppose that does sort of make sense but it was a bit surprising to me.

Yeah, this was surprising to me too, but as far as I can tell from my grepping that's the only usage 🤷 :
```
 (mbta) alerts_concierge :: rg 'route_shapes'
apps/alert_processor/lib/service_info/service_info_cache.ex
655:    {:ok, route_shapes} = ApiClient.route_shapes("Red")
657:    route_shapes

apps/alert_processor/lib/api/api_client.ex
120:  @spec route_shapes(String.t()) :: {:ok, [map]} | {:error, String.t()}
121:  def route_shapes(route) do


 (mbta) alerts_concierge :: rg '/shapes'
apps/alert_processor/lib/api/api_client.ex
122:    "/shapes"

apps/alert_processor/test/fixture/custom_cassettes/service_info.json
<snip>
```

> We do have some /schedules API responses committed to the repo as test fixtures under apps/alert_processor/test/alert_processor/fixture/custom_cassettes (these are used by [ExVCR](https://github.com/parroty/exvcr)), so those might be using the old departure time rules. However, looking at a few of them, it seems like these might not be used anywhere in the current test suite. With maybe some quick automation we could determine this for sure (in order to use a cassette its name must be explicitly written in a use_cassette call). If it turns out we can delete them, we could throw that in as a separate commit within this PR? Or a separate PR if you prefer.
```
 (mbta) alerts_concierge :: rg 'use_cassette "service_info"'
apps/alert_processor/test/alert_processor/service_info/service_info_cache_test.exs
8:    use_cassette "service_info", custom: true, clear_mock: true, match_requests_on: [:query] do
```

It does look like this is in use, but the tests in that file appear to pass? Not sure what to make of that 😕

> It'd be a good verification (and a good exercise for you!) to deploy this branch to dev-green and do an end-to-end smoke test, assuming you now have write access to the Alerts-UI test environment. Let me know if you'd like to pair on this.

Sounds good, are there docs on what this entails somewhere?

apps/alert_processor/lib/service_info/service_info_cache.ex
```diff
@@ -652,14 +658,31 @@ defmodule AlertProcessor.ServiceInfoCache do
...
   defp fetch_route_branches("Red") do
-    {:ok, route_shapes} = ApiClient.route_shapes("Red")
+    {:ok, patterns, included} = ApiClient.route_patterns_with_stops("Red")
 
-    route_shapes
-    |> Enum.reject(fn %{"attributes" => %{"priority" => priority}} ->
-      priority == -1
+    included_by_type =
+      included
+      |> Enum.group_by(fn %{"type" => ty} -> ty end)
```
digitalcora, line 665: suggestion: Stylistically it's more conventional to only use pipelines when more than one pipe operator is involved; otherwise you could just call the function "normally" and save some space. So (with some additional destructuring and use of `&` to reduce variable duplication) this could be condensed to:

```elixir
%{"stop" => stops, "trip" => trips} = Enum.group_by(included, & &1["type"])
stops_by_id = Map.new(stops, &{&1["id"], &1})
trips_by_id = Map.new(trips, &{&1["id"], &1})
```

Or, with more up-front failures, since I did just write that comment about using `fetch!`:

```elixir
%{"stop" => stops, "trip" => trips} = Enum.group_by(included, &Map.fetch!(&1, "type"))
stops_by_id = Map.new(stops, fn %{"id" => id} = stop -> {id, stop} end)
trips_by_id = Map.new(trips, fn %{"id" => id} = trip -> {id, trip} end)
```

apps/alert_processor/lib/service_info/service_info_cache.ex
```diff
@@ -453,7 +453,13 @@ defmodule AlertProcessor.ServiceInfoCache do
            })}
 
         {:ok, schedule} ->
-          [departure_schedule | _t] = Enum.sort_by(schedule, & &1["attributes"]["departure_time"])
+          [departure_schedule | _t] =
+            schedule
+            |> Enum.sort_by(& &1["attributes"]["departure_time"], fn
+              _, nil -> true
+              nil, _ -> false
+              a, b -> a <= b
```
digitalcora, line 461: suggestion: Since we use this exact same logic in two places, maybe it could be extracted to a descriptively-named helper? Then this could be something like:

```elixir
alias AlertProcessor.Helpers.Sort
# ...
[departure_schedule | _t] =
  Enum.sort_by(schedule, & &1["attributes"]["departure_time"], &Sort.nils_last/2)
```

apps/alert_processor/lib/service_info/service_info_cache.ex
```diff
@@ -652,14 +658,31 @@ defmodule AlertProcessor.ServiceInfoCache do
...
+    patterns
+    |> Enum.filter(fn %{"attributes" => %{"typicality" => t}} -> t == 1 end)
+    |> Enum.map(fn %{"relationships" => %{"representative_trip" => %{"data" => %{"id" => id}}}} ->
+      Map.get(trips_by_id, id)
     end)
-    |> Enum.map(fn %{"relationships" => %{"stops" => %{"data" => stops}}} ->
-      Enum.map(stops, & &1["id"])
+    |> Enum.map(fn %{"relationships" => %{"stops" => %{"data" => data}}} ->
+      data
+      |> Enum.map(fn %{"id" => id} -> Map.get(stops_by_id, id) end)
```
digitalcora, line 682: potential issue: These uses of `Map.get` could instead be `Map.fetch!`, which crashes if the key isn't there. It would be surprising if it wasn't, since it would mean the API returned an internally-inconsistent response (or we made a mistake in our code); and in this case we'd probably want to fail rather than try to continue with some unexpected `nil`s thrown in.

digitalcora:
Looking good! I forgot that there should be one more step in testing this on dev-green, and that's to actually set the version of dev-green's API key to the new one. This will be a bit convoluted as it's the first time we've done it in a long time (since... well, 2019 I'd guess), but here's what we need to do:

1. Use the AWS ECS console to look up the API key for the dev-green app (via Secrets Manager)
2. Use the [API admin page](https://api-v3.mbta.com/admin/keys) to look up the account that owns the key
3. Figure out how to access that account — if it's a person we need to ask them to change it, or if it's a shared account there may be credentials somewhere

Let me know if you'd like some help with this. Hopefully the same account owns the keys for all environments, so we only have to do this once.

digitalcora (approved):
Pending a final test on dev-green with the new API version, looks good to go! Let's rewrite this branch with interactive rebase so the "remove unused cassettes" commits are all in one, and the other commits are all in another; then we can "rebase and merge", keeping them as two separate commits on `main`.

PR #1216, opened 2022-04-13
chore: add admin query for invalid email subscription info

zovt:


apps/alert_processor/lib/model/query.ex
```diff
@@ -82,6 +82,20 @@ defmodule AlertProcessor.Model.Query do
...
+      %__MODULE__{
+        label: "Subscription info for users with invalid emails",
+        query: """
+        SELECT
+          u.email AS email,
+          u.communication_mode AS communication_mode,
+          COUNT(DISTINCT s.id) AS total_subscriptions,
+          COUNT(DISTINCT CASE WHEN NOT s.paused THEN s.id END) as unpaused_subscriptions
+        FROM users u
+        INNER JOIN subscriptions s ON u.id = s.user_id
```
digitalcora, line 95: We probably want this to include users with no subscriptions (since we want to see _all_ invalid emails), so this should be a `LEFT JOIN`.

digitalcora:
Aside from one comment, looks good! I'll handle temporarily deploying it to production so we can get the report (need to disable the branch protection for a moment).

digitalcora:
We have the report from this query in production now, so closing without merging.

PR #1217, opened 2022-04-13
API Upgrade `departure_time` handling fixes

zovt:
Figured I'd get this up early for feedback. Still investigating other handling of `departure_time`, but I think this should at least fix the error we were seeing this morning. There's some more information in the first commit here.

I'll keep digging in to this and update this PR with anything else I find


zovt:
I _think_ this is all we need to be `nil`-safe when working with `departure_time` 🤞. The only other place I noticed direct usage of `departure_time` was in `edit.html.eex`, but that is ultimately handed data that passes through `schedule.ex#map_common_trips`, which I believe should already handle the `nil`s properly.

apps/alert_processor/lib/rules_engine/notification_builder.ex
```diff
@@ -4,26 +4,10 @@ defmodule AlertProcessor.NotificationBuilder do
...
-        modified_alert
-
-      {:error, error} ->
-        Logger.warn(fn ->
-          "Error replacing text: alert_id=#{inspect(alert.id)} error=#{inspect(error)}"
-        end)
-
-        alert
-    end
+    do_build_notification({user, subscriptions}, alert)
```
digitalcora, line 10: It seems like the only reason this level of indirection previously existed (`build`/`do_build`) was to perform the text replacement, so we can remove it as well (i.e. cut out lines 10-13).

apps/alert_processor/lib/rules_engine/informed_entity_filter.ex
```diff
@@ -141,7 +141,7 @@ defmodule AlertProcessor.InformedEntityFilter do
   end
 
   defp trip_departure(subscription, schedule) do
-    case Enum.find(schedule, &(&1.stop_id == subscription.origin)) do
+    case Enum.find(schedule, &(&1.stop_id == subscription.origin && &1.departure_time)) do
```
digitalcora, line 144: tiny suggestion: To be more explicit about the type of the value, we could write this condition as

```elixir
&1.stop_id == subscription.origin and not is_nil(&1.departure_time)
```

digitalcora:
> The existing logic when no `stop_id` matches the `subscription_origin` is to return true, so I've extended that to account for the case where there is a matching `stop_id` that has a nil `departure_time`. I'm not 100% sure that this is the desired behavior, but it should be easy enough to adjust if we want this to return false

I think this is acceptable. Zooming out a bit, the purpose of this logic appears to be to make a subscription _not_ match a trip-specific alert if the trip's scheduled departure time from the subscription's origin stop falls outside the user's notification time window. Absent that very specific condition, we default to a match. So I think defaulting to a match is also fine in the case where we can't derive a departure time. (I'm struggling to think of when this would happen in reality, though, since a null departure time normally only happens at the _last_ stop of a trip, which would be unlikely to be a subscription's _origin_... the most likely explanation to me is that the crashes we saw came from "stop subscriptions", where the origin and destination are the same)

> The only other place I noticed direct usage of `departure_time` was in `edit.html.eex`, but that is ultimately handed data that passes through `schedule.ex#map_common_trips`, which I believe should already handle the `nil`s properly.

Unfortunately we might have some more work to do here — looking at that function, I notice this a bit further down from what you changed before:

```elixir
arrival_datetime = NaiveDateTime.from_iso8601!(arrival_timestamp)
departure_datetime = NaiveDateTime.from_iso8601!(departure_timestamp)
```

...which will of course crash if either of those are `nil`.

It's unclear to me exactly what this part of the code is doing, since it starts off with

```elixir
[departure_schedule, arrival_schedule | _] = # ...sort schedules by departure time...
```

which makes no sense to me. Why would you pick the _second_ stop on the trip to determine the trip's overall "arrival time"? Surely it should be the last stop? Some tracing here would be much appreciated 🙇‍♀️

zovt:
> It's unclear to me exactly what this part of the code is doing, since it starts off with
> 
> ```elixir
> [departure_schedule, arrival_schedule | _] = # ...sort schedules by departure time...
> ```
> 
> which makes no sense to me. Why would you pick the _second_ stop on the trip to determine the trip's overall "arrival time"? Surely it should be the last stop? Some tracing here would be much appreciated 🙇‍♀️

So right above the `map` here there's a filter: `Enum.count(schedules) > 1`. As far as I can tell (at least by the test data), the `schedules` we look at are always 2 element lists (maybe we should adjust the aforementioned filter to make that explicit?). There's at least one element with a non-nil `departure_time`, and one with a non-nil `arrival_time` (Sometimes these are separate elements, sometimes there's an element with both a `departure_time` and an `arrival_time`). So with the sort here, before pattern matching we get a list like: `[<always has departure_time>, <always has arrival_time>]`. 

> Unfortunately we might have some more work to do here — looking at that function, I notice this a bit further down from what you changed before:
> 
> ```elixir
> arrival_datetime = NaiveDateTime.from_iso8601!(arrival_timestamp)
> departure_datetime = NaiveDateTime.from_iso8601!(departure_timestamp)
> ```
> 
> ...which will of course crash if either of those are `nil`.

So, assuming the above is true, these will never crash, but this is still pretty fragile. I have no idea if the behavior from above is documented anywhere or reliable. That said, I think we can probably make this more reliable by looking for the schedules with the earliest non-`nil` `departure_time`, and the latest non-nil `arrival_time` and using those as `departure_schedule` and `arrival_schedule` respectively, regardless of the length of `schedules`.

digitalcora:
> As far as I can tell (at least by the test data), the `schedules` we look at are always 2 element lists

Have you found why that might be? Naively I'd expect this to be a response from `/schedules?filter[trip]=...`, which would return the scheduled stop times for all stops on the trip in sequence. I wouldn't be surprised if for the purposes if this function we want it to always be 0-2 stops, but I haven't seen where that happens, in my brief tracing of the code. I agree we should be more explicit about that in the pattern match, and crash if we get something unexpected.

zovt:
> Have you found why that might be? Naively I'd expect this to be a response from `/schedules?filter[trip]=...`, which would return the scheduled stop times for all stops on the trip in sequence. I wouldn't be surprised if for the purposes if this function we want it to always be 0-2 stops, but I haven't seen where that happens, in my brief tracing of the code.

See `schedule.ex:172` and `api_client.ex:137`.

> I agree we should be more explicit about that in the pattern match, and crash if we get something unexpected.

I'll make the changes to make the `arrival_schedule` and `departure_schedule` logic agnostic to the length of the list

digitalcora:
I see, when we fetch `/schedules` in the API client we filter to only schedules for the specified origin and destination stops.

zovt:
Alright, I've added some explicit `raise`-ing if the timestamps are `nil`, let me know if those changes make sense

apps/concierge_site/lib/schedule.ex
```diff
@@ -212,14 +212,11 @@ defmodule ConciergeSite.Schedule do
...
-          fn %{"attributes" => %{"departure_time" => departure_timestamp}} ->
-            departure_timestamp
-          end,
-          Sort.nils_last()
-        )
+      departure_schedule =
+        Enum.min_by(schedules, & &1["attributes"]["departure_time"], Sort.nils_last())
+
+      arrival_schedule =
+        Enum.min_by(schedules, & &1["attributes"]["arrival_time"], Sort.nils_last(&>=/2))
```
digitalcora, line 219: We could use `Enum.max_by` rather than the custom sort function.

apps/concierge_site/lib/schedule.ex
```diff
@@ -240,6 +237,15 @@ defmodule ConciergeSite.Schedule do
       } = departure_schedule
 
       %{"attributes" => %{"arrival_time" => arrival_timestamp}} = arrival_schedule
+
+      if is_nil(departure_timestamp) do
+        raise "departure_timestamp is nil"
+      end
```
digitalcora, line 243: I don't think these explicit raises are needed. The error that will be raised when we try to convert the time string to a time is very clear already:

```
** (FunctionClauseError) no function clause matching in Calendar.ISO.parse_naive_datetime/1

    The following arguments were given to Calendar.ISO.parse_naive_datetime/1:

        # 1
        nil

    Attempted function clauses (showing 1 out of 1):

        def parse_naive_datetime(string) when is_binary(string)
```

digitalcora:
I think we should be all set with this other than my one comment on the `raise`. Could you rebase the commits into their intended final forms, and then I'll approve? ☑️

digitalcora (approved):
Time for round 2!

PR #1218, opened 2022-04-19
chore: add abandonment rate admin query

zovt:
One thing I wasn't 100% sure about: the Asana ticket mentions:

> Should ignore bad emails and any other type of user created by bots, vulnerability scanners, etc.

I checked `email_rejection_status IS NULL` as an attempt to consider this, but I'm not sure if that's the best way to go about it. Let me know if there's something better 👍

### DIFF
--- a/apps/alert_processor/lib/model/query.ex
+++ b/apps/alert_processor/lib/model/query.ex
@@ -82,6 +82,58 @@ defmodule AlertProcessor.Model.Query do
         LEFT JOIN subscriptions s ON u.id = s.user_id
         GROUP BY u.communication_mode
         """
+      },
+      %__MODULE__{
+        label: "Weekly abandonment rate",
+        query: """
+        WITH
+          periods AS (
+            SELECT \"end\" - interval '1 week' AS start, \"end\"
+            FROM generate_series(
+              date_trunc('week', (SELECT MIN(inserted_at) FROM users)) - interval '1 second',
+              date_trunc('week', NOW() AT TIME ZONE 'utc') + interval '1 week' - interval '1 second',
+              interval '1 week'
+            ) \"end\"
+          ),
+          new_users_per_period AS (
+            SELECT COUNT(DISTINCT users.id) AS "count", periods.end AS period_end
+            FROM users
+            INNER JOIN periods ON (
+              periods.start <= users.inserted_at
+              AND users.inserted_at <= periods.end
+            )
+            WHERE
+              users.email_rejection_status IS NULL
+            GROUP BY
+              periods.end
+          ),
+          new_users_without_subscriptions_per_period AS (
+            SELECT COUNT(DISTINCT users.id) AS "count", periods.end AS period_end
+            FROM users
+            INNER JOIN periods ON (
+              periods.start <= users.inserted_at
+              AND users.inserted_at <= periods.end
+            )
+            LEFT JOIN subscriptions ON subscriptions.user_id = users.id
+            WHERE
+              users.email_rejection_status IS NULL
+              AND subscriptions.id IS NULL
+            GROUP BY
+              periods.end
+          )
+        SELECT
+          to_char(periods.end, 'YYYY-MM-DD') AS period_end,
+          COALESCE(new_users_per_period.count, 0) AS new_users,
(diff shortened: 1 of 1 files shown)


digitalcora:
Hmm, I realize now the Asana task may not have been perfectly clear on this... but the idea of these queries is that they would be for _all_ of the specified time periods (for which data exists), not just the most recent one, as we want to leverage all the historical data we've already stored. So the result of this query should be a table like:

week_starting | without_subscriptions | with_subscriptions
-- | -- | --
2022-04-17 | 17 | 32
2022-04-10 | 4 | 15
2022-04-03 | 9 | 24
... | ... | ...

With this in mind, you might want to take a second look at whether `CROSS JOIN` will be performant in all the places you're using it (we don't care about performance _that_ much for an infrequently-run admin query, but we do want to avoid loading down the production database server for a significant time).

> I checked `email_rejection_status IS NULL` as an attempt to consider this

Yep, I think this should be fine.

zovt:
Here's an example of what this looks like. The `abandonment_rate` column will be `null` when we haven't had any user signups

digitalcora:
Nice, looking good! Few general things for polish:

1. It looks like the `period_end` is at the end of the day on Monday, which means the "weeks" in the report start on Tuesday. Probably they should start on Sunday. This also appears to be in UTC, but we'd want the day boundaries to be in Eastern time if that's easily possible. (not essential, given our offset from UTC isn't that far and these are weekly stats rather than daily)

2. It'd be nice if we could format the `period_end` as just an ISO date; the extra `23:59:59` is just noise.

3. Rather than 100 weeks, can we have this go all the way back to when the first-ever user was inserted? I'll do a test deploy of this branch to production before we merge it to make sure the performance is okay, but in the abstract it feels like if it's fine at 100 weeks it should be fine at ~500 (and 100 weeks sadly isn't even enough to get us out of Pandemic Time™ and look at how things might have changed before and after that transition).

4. I'd rather not include derived stats like `abandonment_rate` in reports when they can be calculated entirely from other data that's already there — we have the CSV export feature, so it's pretty easy to import that into a spreadsheet and add a calculated column if desired.

digitalcora:
Did a test deploy and run of this query on production. It only took about 1 second, so I think for now it's fine if we expand the time range to all users. Long-term, if we keep wanting more complex admin queries, I'd like to build in a feature where queries can have "inputs", so like for this one you could specify the date range you want (and then if you only wanted a small date range it would be much faster).

apps/alert_processor/lib/model/query.ex
```diff
@@ -82,6 +86,58 @@ defmodule AlertProcessor.Model.Query do
...
         """
+      },
+      %__MODULE__{
+        label: "Abandonment rate in the past 500 weeks",
+        query: """
+        WITH
+          periods AS (
+            SELECT \"end\" - interval '1 week' AS start, \"end\"
+            FROM generate_series(
+              #{@nearest_weekly_period_end} - interval '500 weeks',
```
digitalcora, line 97: Instead of a fixed interval, can we put the result of `SELECT MIN(inserted_at) FROM users` here? That way it will always go back for as long as we have data, and also won't go any further than it needs to (testing this on production right now, there's a whole bunch of extra rows at the end that are all zeros).

apps/alert_processor/lib/model/query.ex
```diff
@@ -82,6 +82,58 @@ defmodule AlertProcessor.Model.Query do
...
+              AND users.inserted_at <= periods.end
+            )
+            LEFT JOIN subscriptions ON subscriptions.user_id = users.id
+            WHERE
+              users.email_rejection_status IS NULL
+              AND subscriptions.id IS NULL
+            GROUP BY
+              periods.end
+          )
+          SELECT
```
digitalcora, line 124: Minor indentation issue: I think this line and all the following ones should be de-indented one level (as this is the end of the `WITH` clause that introduced the extra level of indentation).

digitalcora (approved):
Other than one tiny comment, looks great! I deployed this version to prod and confirmed it works fine across the whole time range.

apps/alert_processor/lib/model/query.ex
zovt: 👀

PR #1220, opened 2022-04-21
chore: add queries to list and delete users with invalid emails

zovt:


zovt:
The unit test that is failing is presumably to prevent mutative queries from being present as an admin query. Not sure if just ignoring it for now is valid, if not I can temporarily disable it or add some sort of filtering for specifically-allowed queries if this is a pattern we want to adopt more in the future.

digitalcora:
Since we don't plan to merge this work to the main branch, I believe it's fine to ignore the unit test error, as it doesn't block us from temporarily deploying to production (which I'm doing now). Excellent idea making it two queries to be able to preview the affected users before running.

digitalcora:
All set with this! Delete query has been run on production. Closing without merging.

digitalcora:
This should now be all set. I threw in a `LIMIT 50` on the delete query due to hitting some timeouts.

PR #1221, opened 2022-04-21
chore: add queries to list and fix invalid emails

zovt:
I _think_ this should cover most of the invalid emails, but probably worth deploying and iterating on the replacements if we notice anything missing

### DIFF
--- a/apps/alert_processor/lib/model/query.ex
+++ b/apps/alert_processor/lib/model/query.ex
@@ -49,6 +49,7 @@ defmodule AlertProcessor.Model.Query do
     FROM users
     WHERE
       users.email !~ '^[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,64}$'
+      AND users.email != #{email_replacements()}
       AND #{email_replacements()} IN (SELECT email FROM users)
       #{additional_where_clause}
     """
--- a/apps/alert_processor/lib/model/query.ex
+++ b/apps/alert_processor/lib/model/query.ex
@@ -19,6 +19,29 @@ defmodule AlertProcessor.Model.Query do
 
   @users_with_subscriptions "users u JOIN subscriptions s ON u.id = s.user_id"
 
+  defp email_replacements(replacements) do
+    Enum.reduce(replacements, "users.email", fn {l, r}, acc ->
+      "regexp_replace(#{acc}, '#{l}', '#{r}')"
+    end)
+  end
+
+  defp email_replacements() do
+    email_replacements([
+      {"[ !,]", ""},
+      {"@@*", "@"},
+      {"(@gma$)|(@gmil.com)|(@g$)|(@gmail.co$)", "@gmail.com"},
+      {"@aol$", "@aol.com"}
+    ])
+  end
+
+  defp select_invalid_emails(select_clause) do
+    """
+    SELECT #{select_clause}
+    FROM users
+    WHERE users.email !~ '[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,64}'
+    """
+  end
+
   @doc "Lists available queries."
   @spec all() :: [t]
   def all do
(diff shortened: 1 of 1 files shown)


digitalcora:
One comment: as we discussed earlier, let's set `digest_opt_in = false` for users whose emails we fix, since if they had tried to opt in before, Mailchimp likely would have rejected their email. Otherwise this looks good!

zovt:
> One comment: as we discussed earlier, let's set `digest_opt_in = false` for users whose emails we fix, since if they had tried to opt in before, Mailchimp likely would have rejected their email. Otherwise this looks good!

Ah yep 🤦‍♂️ thanks for the reminder

digitalcora:
Ran the SELECT query on production. Some observations:

1. The regex replacement seems like it isn't "global" — for example there are some emails with multiple space characters in them, and only the first is removed.
2. There are three emails where the domain part is just `@com`, with no clue what the actual domain might have been. Likewise there is one that's just `@thl`. According to the report, all have active subscriptions, so I'm not sure what to do with these. The only thing I can think of is that maybe we can enhance the query to check their SMS opt-out status? If they are in fact opted out, they're not receiving notifications anyway, and so are in the "ok to clean up" category.

zovt:
> The regex replacement seems like it isn't "global" — for example there are some emails with multiple space characters in them, and only the first is removed.

Ah yeah, good catch, I'll fix that

> There are three emails where the domain part is just @com, with no clue what the actual domain might have been. Likewise there is one that's just @thl. According to the report, all have active subscriptions, so I'm not sure what to do with these. The only thing I can think of is that maybe we can enhance the query to check their SMS opt-out status? If they are in fact opted out, they're not receiving notifications anyway, and so are in the "ok to clean up" category.

I think if they were not receiving any alerts, they would have been cleaned up with the deletion query no?

digitalcora:
> I think if they were not receiving any alerts, they would have been cleaned up with the deletion query no?

Ah, right, I forgot that opting out of SMS also sets your communication mode to `none`. Well, then I'm not quite sure what to do with these ones. I don't think there is anything we _can_ do programmatically other than delete them, which may not be desirable... something to talk about with Steve next time.

digitalcora:
I got a "duplicate key" error running the fix query on production, which I guess isn't surprising — if someone didn't remember they had typo'd their email when initially creating an account, they might have ended up creating another one with the correct email. In these cases, I think it's probably fine to delete the account with the typo'd email; we might need another query to do that, before running this one. (it can be part of this branch)

digitalcora:
Ran the updated "list fixes" query and attached the result to the subtask in Asana, so you can get an idea of what else might be fixable programmatically.

Also ran the "list duplicates" query, but in some cases the automatic fixes didn't do anything, and the address in the left and right columns was the same. I think this means the real delete query would delete accounts as "duplicates" when it shouldn't, since the "duplicate" it found is just the same account? Might want to create some local users to test that (either manually or automated as a test case, if that helps).

digitalcora:
Tried again on production. "Delete duplicate fixed emails" doesn't seem to do anything, and returns 0 rows.

zovt:
> Tried again on production. "Delete duplicate fixed emails" doesn't seem to do anything, and returns 0 rows.

I just ran this again locally and confirmed that it did delete the emails that the list query says will be deleted. Did all of the results have `will_be_deleted = false`?

digitalcora:
Ah, you're right, I didn't notice the addition of that column — they are all `will_be_deleted = false`. But that doesn't seem to make sense, as for most of them, the "duplicate" email _is_ different (for example it has a space on the end), so it should be deleted. 🤔

zovt:
> Ah, you're right, I didn't notice the addition of that column — they are all `will_be_deleted = false`. But that doesn't seem to make sense, as for most of them, the "duplicate" email _is_ different (for example it has a space on the end), so it should be deleted. 🤔

Are they all using `sms`? I think we decided that we'll keep "duplicate" accounts where `sms` is the communication mode

zovt:
or rather, we only want to delete accounts where `communication_mode = "email"`

digitalcora:
I see — I think I was assuming we'd even delete those, but now that you mention it we should probably get a Steve ruling on this. It certainly ends up leaving a lot more invalid emails in the database than I'd like.

Anyway, fixable emails have now been fixed on production, so I will close this out pending discussion about how to handle the remaining invalid emails (including those that aren't fixable at all).

PR #1222, opened 2022-04-22
fix: make email validation regex match full text

zovt:


digitalcora:
@zovt Looks good, but could we add some test coverage? Like testing that a user with email `bad email@example.com` is not considered valid.

apps/alert_processor/test/alert_processor/model/user_test.exs
```diff
@@ -142,6 +142,10 @@ defmodule AlertProcessor.Model.UserTest do
       refute create_changeset(%{"email" => "email @example.com"}).valid?
     end
 
+    test "is invalid when whole string is not an valid email" do
```
digitalcora, line 145: small typo: "an valid" => "a valid"

digitalcora (approved):
One teensy comment. Thanks for the extra PR, this will definitely be helpful to prevent some kinds of typos when folks are signing up.

digitalcora:
Looks like you've found your first live-data test failure 😅

Side note, I like to use the conventional commit prefix `test:` for commits that exclusively change/add/remove tests; [here's one example](https://github.com/mbta/alerts_concierge/commit/ab7cc9c8224cd0c26ae2fc0131201a52413374b8). Other good prefixes to use are `ci:` and `docs:` for commits that only change CI workflows and documentation, respectively, and `refactor:` for commits that change code but don't change any app behavior. I find these improve the "skimmability" of the Git history, personally, but it's not at all essential.

PR #1223, opened 2022-04-26
chore: add retention rate query

zovt:
Example output:

These columns match what was requested in the ticket, though maybe we can drop the `start` column and just have `new` and `end`? 🤔

### DIFF
--- a/apps/alert_processor/lib/model/query.ex
+++ b/apps/alert_processor/lib/model/query.ex
@@ -134,6 +134,55 @@ defmodule AlertProcessor.Model.Query do
           ON new_users_without_subscriptions_per_period.period_end = periods.end
         ORDER BY periods.end DESC
         """
+      },
+      %__MODULE__{
+        label: "Monthly new active users",
+        query: """
+        WITH
+          periods AS (
+            SELECT start, start + interval '1 month' - interval '1 day' as \"end\"
+            FROM generate_series(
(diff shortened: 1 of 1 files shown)


digitalcora:
Certainly it seems like only two of the three columns should be required, as the third can then be derived — I'd go with whatever is easiest in the SQL. We should probably also make the value `0` instead of `null` where applicable, otherwise we'd have to manually correct those values if loaded into a spreadsheet or other program.

Just looking at that screenshot, it seems like there's an issue with the dates... after February, the end of each period is the 28th of the month, but the following months have more than 28 days.

zovt:
> Just looking at that screenshot, it seems like there's an issue with the dates... after February, the end of each period is the 28th of the month, but the following months have more than 28 days.

Yeah... I think it's probably related to how Postgres handles monthly intervals. I'll fiddle with it and see if I can peg it to the end of each month easily

digitalcora:
This looks good, though I think we should be a bit clearer about the meaning of the columns, since there is some complexity hidden in there. Maybe `new_active_users` and `total_active_users`? (we already use the term "active" in other reports to describe users who have at least one unpaused subscription)

Also, "retention rate" seems like... maybe not the right name for what this report is, vs. the original intent? I'm not sure what a better name would be. Maybe "Monthly new active users"?

digitalcora (approved):
📆 👥 ✅

PR #1225, opened 2022-04-28
feat: add tracking pixel to notification email and log opening

zovt:


apps/concierge_site/lib/helpers/mail_helper.ex
```diff
@@ -47,4 +47,12 @@ defmodule ConciergeSite.Helpers.MailHelper do
   def reset_password_url(reset_token) do
     Helpers.password_reset_url(ConciergeSite.Endpoint, :edit, reset_token)
   end
+
+  @spec track_open_url(%{alert_id: String.t(), notification_id: String.t()}) :: iodata
+  def track_open_url(%{alert_id: alert_id, notification_id: notification_id}) do
+    Helpers.static_url(
+      ConciergeSite.Endpoint,
+      "/notification_email_opened?alert_id=#{alert_id}&notification_id=#{notification_id}"
```
digitalcora, line 55: Having this defined in the Router generates a `notification_email_opened_url` helper that we should probably use here if we can, instead of manually constructing the path with `static_url`.

apps/concierge_site/lib/mail_templates/notification.html.eex
```diff
@@ -344,5 +344,6 @@
 
       </div>
 
+      <img src="<%= track_open_url %>" height=0 width=0>
```
digitalcora, line 347: We should also add this to the original MJML template so we don't forget about it when we do that cleanup.

apps/concierge_site/lib/controllers/email_opened_controller.ex
```diff
@@ -0,0 +1,10 @@
+defmodule ConciergeSite.EmailOpenedController do
+  use ConciergeSite.Web, :controller
+  require Logger
+
+  def notification(conn, %{"alert_id" => alert_id, "notification_id" => notification_id}) do
+    Logger.info("email opened alert_id=#{alert_id} notification_id=#{notification_id}")
+    conn = put_resp_content_type(conn, "image/gif", nil)
+    send_resp(conn, 200, "")
```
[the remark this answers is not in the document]
zovt: it is not

apps/concierge_site/lib/helpers/mail_helper.ex
```diff
@@ -47,4 +47,12 @@ defmodule ConciergeSite.Helpers.MailHelper do
   def reset_password_url(reset_token) do
     Helpers.password_reset_url(ConciergeSite.Endpoint, :edit, reset_token)
   end
+
+  @spec track_open_url(%{alert_id: String.t(), notification_id: String.t()}) :: iodata
+  def track_open_url(%{alert_id: alert_id, notification_id: notification_id}) do
```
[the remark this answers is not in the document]
zovt: Yeah this is mostly just a personal preference to help avoid mixing up parameter order, but I can stick to regular arguments if this is too strange

apps/concierge_site/lib/mail_templates/notification.html.eex
zovt: ah, good call out 👍 thanks!

apps/concierge_site/lib/helpers/mail_helper.ex
```diff
@@ -47,4 +47,14 @@ defmodule ConciergeSite.Helpers.MailHelper do
   def reset_password_url(reset_token) do
     Helpers.password_reset_url(ConciergeSite.Endpoint, :edit, reset_token)
   end
+
+  @spec track_open_url(%{alert_id: String.t(), notification_id: String.t()}) :: iodata
```
digitalcora, line 51: The return type of URL helpers is `String.t()`, rather than the more general `iodata`.

apps/concierge_site/lib/helpers/mail_helper.ex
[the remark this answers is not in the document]
digitalcora: I'd say it's "unconventional", in the sense that someone familiar with general Elixir conventions might wonder if there's a special reason why it was done this way. Not really all that strange, though.

An alternative, if you'd like, is to pass the whole `Notification`, simplifying the call while maintaining the destructuring pattern. The helper itself doesn't even need to know it's being passed a Notification, and _can_ still be called with an arbitrary map, e.g. for testing.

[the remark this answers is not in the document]
zovt: 🤔 I think I'll try passing in the `Notification`, thanks for the suggestion 👍

apps/concierge_site/lib/helpers/mail_helper.ex
zovt: 🤔 got it, is there a reason that dialyzer doesn't catch this?

digitalcora (approved):
Looks good, pending remaining comments being resolved one way or the other!

apps/concierge_site/lib/helpers/mail_helper.ex
digitalcora: By default it's not an error if an explicit type is a superset of the success type; you need to enable the optional `underspecs` flag, which we haven't on this project. Trying that out might be a good idea, if it doesn't produce many errors on the existing code.

zovt: cool, good to know

PR #1226, opened 2022-04-29
fix: add overload to catch missing query params

zovt:
Forgot to consider that the endpoint could potentially be hit without the query params set 🥲 

currently this just swallows the request, but do we want to log anything in this case? I think maybe this could have been originally caused by email clients scrubbing out query parameters on images 🤔 could be an interesting stat to track

### DIFF
--- a/apps/concierge_site/lib/helpers/mail_helper.ex
+++ b/apps/concierge_site/lib/helpers/mail_helper.ex
@@ -51,11 +51,6 @@ defmodule ConciergeSite.Helpers.MailHelper do
 
   @spec track_open_url(Notification.t()) :: String.t()
   def track_open_url(%Notification{id: notification_id, alert_id: alert_id}) do
-    Helpers.email_opened_url(
-      ConciergeSite.Endpoint,
-      :notification,
-      alert_id: alert_id,
-      notification_id: notification_id
-    )
+    Helpers.email_opened_url(ConciergeSite.Endpoint, :notification, alert_id, notification_id)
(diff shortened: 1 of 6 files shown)


zovt:
One other thing I noticed while splunking: searching for `email_opened` also includes some HTTP logging for hitting the `notification_email_opened` endpoint. I can narrow it down to the proper events by searching for `" email_opened "`, but I wonder if it's worth changing `email_opened` to something distinct from the endpoint

digitalcora:
> I wonder if it's worth changing `email_opened` to something distinct from the endpoint

I think so, yes. Either that or we could change the route to something distinct.

apps/concierge_site/lib/helpers/mail_helper.ex
```diff
@@ -54,8 +54,8 @@ defmodule ConciergeSite.Helpers.MailHelper do
     Helpers.email_opened_url(
       ConciergeSite.Endpoint,
       :notification,
-      alert_id: alert_id,
-      notification_id: notification_id
+      alert_id,
+      notification_id
```
digitalcora, line 58: I think this all fits on one line now. `mix format` never removes line breaks, even if doing so would allow fitting a statement on fewer lines.

digitalcora (approved):
One small suggestion, looks good!

digitalcora:
Oh, and after you deploy this to production, don't forget to mark the Sentry issue as resolved ☑️

PR #1227, opened 2022-05-02
chore: clean up unnecessary email handling code

zovt:
Following some clean up work done recently, all of our emails should be "valid" now (in the sense that the should at least be well-formed, not necessarily real).

Once this merges in, I'll keep my eye on Sentry for any new errors that pop up


arkadyan (approved):
👍

PR #1228, opened 2022-05-03
chore: track notification batches

zovt:
We want to track some operational performance metrics in Splunk. One thing that we want to track is the total time taken to send all notifications for an alert. However, since alerts can persist over longer periods of time, tracking just the `alert_id` is insufficient. Thus, with this PR we want to add a concept of a `batch_id`. This will allow us to see how long it takes for us to send all notifications for an alert _in a given batch_ rather than over the entire lifetime of the alert.


arkadyan (approved):
👍 Nice, I like this solution. Pinging @ryan-mahoney just because he might find it interesting as well. At some point, we were thinking through ways to handle this very problem, and I think maybe we were considering trying to log how long since the alert "ideally should have been sent", or something like that, but this approach seems much more straightforward.

PR #1229, opened 2022-05-05
Use MJML templates directly

zovt:
Currently, to update the HTML content of an email the workflow looks like this:
1. update the MJML template
2. manually convert the MJML to HTML using an external tool
3. update the related cached file in `mail_templates/`

After this, we can just do step 1.

Some things to note:
- I also cleaned up some other unused stuff, including most of the files in `mail_templates/` and some functions in `mail_helper.ex`. I'm pretty confident that these were not referenced anywhere, but it could probably use another pair of eyes. Regardless, I'll keep an eye on Sentry after deploying this.
- I added a comment to silence `sobelow` in `mjml.ex`. Based on our current usage of the module, the error that it finds is a false positive, but there is a possibility that functions in that module could be abused if we're not careful. Not sure if there's much we can do, maybe a comment describing the safety conditions somewhere in the module could be useful? Definitely open to suggestions here!


apps/concierge_site/lib/helpers/mjml.ex
```diff
@@ -0,0 +1,93 @@
...
+  @spec to_html_with_eex(String.t()) :: {:ok, String.t()} | {:error, any()}
+  def to_html_with_eex(contents) do
+    case Mjml.to_html(escape_eex(contents)) do
+      {:ok, contents} -> {:ok, unescape_eex(contents)}
+      e -> e
+    end
+  end
+
+  @spec html_from_template(Path.t()) :: {:ok, String.t()} | {:error, any()}
+  def html_from_template(name) do
```
arkadyan, line 40: question: Could this function be private? (Same with, e.g., `to_html_with_eex`.)

apps/concierge_site/lib/helpers/mjml.ex
```diff
@@ -0,0 +1,93 @@
...
+  def eval_string(mjml_string, bindings \\ [], options \\ []) do
+    case to_html_with_eex(mjml_string) do
+      # sobelow_skip ["Traversal"]
+      {:ok, html} -> {:ok, EEx.eval_string(html, bindings, options)}
+      e -> e
+    end
+  end
+
+  @spec function_from_template(:defp | :def, atom(), Path.t(), [atom()], keyword) :: Macro.t()
+  defmacro function_from_template(kind, name, template, args \\ [], options \\ []) do
```
arkadyan, line 70: suggestion: Since this is the primary public "interface" to this module, it seems like it would be nice to have at least a quick test demonstrating the functionality.

apps/concierge_site/test/lib/helpers/mjml_test.exs
```diff
@@ -0,0 +1,23 @@
+defmodule ConciergeSite.MJMLTest do
+  @moduledoc false
+  use ConciergeSite.DataCase, async: true
+  alias ConciergeSite.MJML
+
+  test "works properly with EEx tags" do
```
arkadyan, line 6: nitpick: This could be wrapped with a `describe` block saying that this is testing `eval_string`.

apps/concierge_site/assets/mjml/confirmation.mjml
```diff
@@ -67,7 +67,7 @@
 
     <mj-section background-color="#ffffff" padding-top="0" padding-bottom="20">
       <mj-column>
-        <mj-image href="<%= manage_subscriptions_url %" src="https://s3.amazonaws.com/mbta-alerts/html_email/welcome/sign-in-button.png" width="189" height="45"></mj-image>
+        <mj-image href="<%= manage_subscriptions_url %>" src="https://s3.amazonaws.com/mbta-alerts/html_email/welcome/sign-in-button.png" width="189" height="45"></mj-image>
```
arkadyan, line 70: 🙀

arkadyan (approved):
👍 Nice. I added a few tiny comments, but I think the `function_from_template` test suggestion is the one that feels most important to me.

apps/concierge_site/test/lib/helpers/mjml_test.exs
zovt: this is using `eval_string`, but the main concept that I wanted to test here is that escaping/unescaping EEx works and that it ultimately gets evaluated properly which feels to me like it is more related to the module as a whole rather than just `eval_string` 🤔 e.g. if we added something like `eval_file` in the future I don't think we should duplicate a similar test under a separate `describe "eval_file/3"` block, if that makes sense?

arkadyan: Ah yeah, sure, that makes sense.

PR #1230, opened 2022-06-13
Fix Failing Tests

zovt:
Fixes a couple of test failures due to live data changes


arkadyan (approved):
👍 

Interesting that that one station moved.

PR #1231, opened 2022-06-13
fix: manually rate limit fetching opted out list in opt out worker

zovt:
Let me know if there are any concerns here around using `Process.sleep` that I'm not aware of :)

### DIFF
--- a/apps/alert_processor/lib/dissemination/sms_opt_out_worker.ex
+++ b/apps/alert_processor/lib/dissemination/sms_opt_out_worker.ex
@@ -67,8 +67,15 @@ defmodule AlertProcessor.SmsOptOutWorker do
         normalized_phone_numbers = Enum.map(phone_numbers, &String.replace_leading(&1, "+1", ""))
 
         case next_token do
-          "" -> {:ok, opted_out_list ++ normalized_phone_numbers}
-          token -> fetch_opted_out_list(token, opted_out_list ++ normalized_phone_numbers)
+          "" ->
+            {:ok, opted_out_list ++ normalized_phone_numbers}
+
+          token ->
+            # The underlying API call here has a hard rate limit of 10 requests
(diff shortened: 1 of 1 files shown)


zovt:
Looks like it ran to completion on dev-green 👍

zovt:
Just checked back on this this morning. We're still getting occasional failures, so I'm backing the timeout off again

digitalcora:
Looking at the dev-green logs now, it looks like it mostly succeeds, but still fails due to "throttling" about a quarter of the time, despite the other instance logging `skipped` indicating it is respecting the lock.

It occurs to me that the 10 requests/sec limit is likely at the level of the AWS _account_, meaning it's actually shared among all three `alerts_concierge` environments (prod, dev, dev-green). Since the check is not clock-scheduled but instead occurs 1 hour after the completion of the last check, the check times for each environment will drift around and sometimes overlap.

I think this is probably okay for a short-term fix, since it should still succeed every few hours, and there is no downside (other than the expense) to sending messages to an opted-out number for a while — they are blocked at the carrier level immediately once the stop keyword is sent.

If we wanted to really make sure we stop sending ASAP, we could set the delay to >300ms, which would allow the checks in all three environments to be running concurrently; with the current list size, each check would "only" take ~45s. Not a huge fan of this, considering it's tying an aspect of the code to the number of environments it happens to be deployed to, but again, might be acceptable if we plan to delete this soon anyway. Your call!

zovt:
Hah that all makes sense :) I think for now sticking with a 200ms timeout should be a slightly better balance between success and failures. I'm leaning towards axing this code anyways once we start receiving the SMS delivery failures

digitalcora:
Also, we should expect to still see some failures on dev-green even if the delay was ideal, because this same code is not (yet) deployed to dev or prod, and so those environments are going to be less respectful of the rate limit.

PR #1232, opened 2022-06-27
SMS Logs: Initial Lambda Code

zovt:
# Overview
This is the initial code for the SMS log-forwarding Lambda. It is set up as a separate Node.js project in the `sms_logs` directory. Eventually this can be extended to replace the existing Lambda that is forwarding our logs to Splunk, but for now this only handles forwarding them to T-Alerts.

The basic idea is that we'll forward both the number which we got an error about as well as the error message itself. Each request also includes a `secret` token that will be stored in AWS to prevent any unauthorized usage of the (not yet implemented on the server side) hook.

# Questions
- Do we want to forward _all_ logs to T-Alerts? We currently have a separate Splunk index specifically for the SMS logs, but this could be a good opportunity to re-evaluate that and put everything in the T-Alerts indices if there is value in that (e.g. Splunk queries we want to write but currently can't because the data is stored across indices).
- Do we want to add a step to the GitHub CI to run the `check` step for this code?


digitalcora:
> Do we want to add a step to the GitHub CI to run the `check` step for this code?

Yes, definitely. In fact I think I'd prefer to implement `check.sh` as an Actions workflow instead of a shell script, since then we can perform the different checks in parallel and have independent reports of their status (easy to see you left in a formatting issue when the "Format" step has an ❌ and the others have ✅). If we had this, I'm not sure if the shell script would have much value on its own, at least not enough to justify keeping the steps in sync across two locations.

I think the way to handle this might be to make an entirely new "Lambda CI" workflow, separate from the existing "CI" workflow. We can use `paths` to only run the new workflow on changes to files within `sms_logs`, and also add that directory to `paths-ignore` in the existing "CI" workflow. The advantage here is that it will be a bit easier to set up than putting everything in "CI", but the disadvantage is that if you're changing both the main app and the Lambda code at the same time, the CI results will be in two different places.

Does that make sense? Let me know if anything about how to set up the workflow is unclear. You should be able to test a new workflow just by pushing it to this branch.

sms_logs/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "jest": "^28.1.1",
+    "prettier": "^2.7.1",
+    "ts-jest": "^28.0.5",
+    "typescript": "^4.7.3"
+  },
+  "dependencies": {
+    "@aws-sdk/client-secrets-manager": "^3.112.0",
+    "@sentry/serverless": "^7.2.0",
+    "node-fetch": "^3.2.6",
+    "zod": "^3.17.3"
```
digitalcora, line 33: Not that I have strong opinions about the specific choice of library (they seem quite similar), but we're currently using `superstruct` for this same purpose in OCS Saver, and I do have opinions about sticking to the same library for the same job across projects 🙂 Any compelling reason we should prefer this one?

Also an option: wait a few days, and you can ignore all of my opinions 😄

sms_logs/src/alerts.test.ts
```diff
@@ -0,0 +1,106 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
digitalcora, line 1: question: We don't seem to have an explicit `any` in this file. What is this needed for?

zovt:
> Yes, definitely. In fact I think I'd prefer to implement `check.sh` as an Actions workflow instead of a shell script, since then we can perform the different checks in parallel and have independent reports of their status (easy to see you left in a formatting issue when the "Format" step has an ❌ and the others have ✅). If we had this, I'm not sure if the shell script would have much value on its own, at least not enough to justify keeping the steps in sync across two locations.

Yep, makes sense to me (though fwiw `check.sh` does run each check in parallel and report which ones failed 🙂)

> I think the way to handle this might be to make an entirely new "Lambda CI" workflow, separate from the existing "CI" workflow. We can use `paths` to only run the new workflow on changes to files within `sms_logs`, and also add that directory to `paths-ignore` in the existing "CI" workflow. The advantage here is that it will be a bit easier to set up than putting everything in "CI", but the disadvantage is that if you're changing both the main app and the Lambda code at the same time, the CI results will be in two different places.

Oooh interesting, didn't know about that. This sounds good, I'll mess around with it a bit and come back with any questions.

sms_logs/src/aws.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  return logData;
+}
+
+export async function createAlertsConfig() {
+  const client = new SecretsManager({ region: "us-east-1" });
+
+  const endpoints = await Promise.all(
+    ["alerts", "alerts-dev", "alerts-dev-green"].map(async (s) => {
+      const url = `https://${s}.mbtace.com/sms_failure`;
```
digitalcora, line 19: The canonical domain for production is `alerts.mbta.com` — though this technically also works, I'm not sure we should rely on it.

Backing up a bit, shouldn't we configure the URL in the environment, and have (at least) separate "dev" and "prod" instances of this Lambda? Then we can test out changes in a deployed environment, before deploying them to production. (For this PR we don't have to think about the details of the environments; this script can just assume there will be an environment variable that contains the URL to report to.)

sms_logs/src/aws.ts
```diff
@@ -0,0 +1,40 @@
...
+      const secretName = `${s}-sms-failure-key`;
```
digitalcora, line 20: Can we make the secret itself available in an environment variable? I know we do this with many of our ECS apps, for example, and that we've been trying to move to that approach over using Secrets Manager as it's much simpler to implement on the app side.

digitalcora (commented):
If feasible, I think it'd be a good idea to get the infra side of this in place (at least a "dev environment" for the Lambda) and test the code there before this is merged — otherwise there might be some churn as we realize what we committed the first time doesn't quite work. Basically set up Terraform with the initial "dummy" Lambda, get that merged and applied, then we have a place to deploy this code and confirm it works for real. (still some risk of churn if the infra itself doesn't work how we wanted it, but idk, gotta start somewhere)

sms_logs/package.json
zovt: Nope, `zod` was just one I'm familiar with. Happy to substitute with `superstruct` for consistency 👍

sms_logs/src/alerts.test.ts
zovt: Probably just left over from some changes

sms_logs/src/aws.ts
zovt: hmmm... so the trick here is that I believe we can only have 2 "subscription filters" per log group. We're using up one of the slots for the existing Splunk forwarding, and the plan was for this lambda to use the other one (so we'd be temporarily maxing them out until this replaces the Splunk forwarding).

Given the above, I think we could go with an env var approach though it would get a bit messy. Any other thoughts on this?

sms_logs/src/aws.ts
zovt: Similar concerns here regarding using env vars.

sms_logs/src/aws.ts
digitalcora: Then this limit seems like a blocker for being able to test new versions of the Lambda code in a production-like environment without actually deploying to production, which normally we place a pretty high value on. I will defer to @ianwestcott on this since I'm on the way out, but my initial reaction would be to look into alternate implementations that would allow us to maintain a dev/prod separation. (Ideally we'd have done so before writing all this code... and I'll definitely take the blame for that, since I approved of this approach in our earlier meeting without fully understanding the implications of the 2-filter limit.)

Our options are probably:

1. Proceed with the current Lambda approach, and just be really careful about testing and deploying changes in the future, since our only option is to go straight to production. In this case I'd recommend configuring the Lambda using an environment variable that holds a JSON-encoded object, which can then be as complex as needed to provide the multiple reporting URLs and corresponding secret names (the Lambda code would still have to use Secrets Manager to look up the values). Terraform has a `jsonencode` function which makes this very easy, and there is precedent with [Concentrate](https://github.com/mbta/devops/blob/master/terraform/prod/concentrate.tf).

2. Hook up the subscription filter to a Kinesis stream (which seems to be the only other option besides a Lambda). This allows using Firehose delivery streams to send the events directly to HTTP endpoints of our choosing, meaning no Lambda code would be required. There would be separate delivery streams per T-Alerts environment, but they could all pull from the same single Kinesis stream. Firehose does support configuring an arbitrary "access key" which is sent to the endpoint as an `X-Amz-Firehose-Access-Key` header, [as documented here](https://docs.aws.amazon.com/firehose/latest/dev/httpdeliveryrequestresponse.html), so that would cover the authorization piece. There is some additional infra complexity due to the requirement to set up an S3 bucket/prefix where events go if HTTP delivery fails (you can't just drop them on the floor, even though we might be fine with that here).

Notably, Firehose also supports delivery to Splunk. However, the extra infra complexity and the fact that there's an account-wide soft limit of 50 Firehose delivery streams suggests that maybe we shouldn't aim to replace our existing Lambda-based Splunk forwarding with this approach.

zovt:
We're going to close this out. We've decided that for now, the `SMSOptOutWorker` is enough to meet our current needs.

PR #1234, opened 2022-07-19
feature: use `check_if_phone_number_is_opted_out` in `SMSOptOutWorker`

zovt:
# Overview
This PR reworks the opt out worker logic to check phone numbers in the database against the AWS opted-out list using `check_if_phone_number_is_opted_out`, rather than pulling the entire opted-out list via `list_phone_numbers_opted_out`.

Also fixes a live data failure

## Things to Note
- The `ExAws.SNS` version we were using doesn't support the `check_if_phone_number_is_opted_out` call, so I've upgraded to latest git version. Once they tag an official release, I'll switch to using it instead of the git ref.
- The previous logic would fail the entire run if an error occurred. I've adjusted the logic here to instead continue on error, logging and incrementing an error count.

# Testing Done
I've added some unit tests to check the functionality as well as tested on dev-green using my own number.


apps/concierge_site/mix.exs
```diff
@@ -41,7 +41,8 @@ defmodule ConciergeSite.Mixfile do
       {:bamboo, "~> 2.2.0"},
       {:bamboo_phoenix, "~> 1.0.0"},
       {:bamboo_ses, "~> 0.3.0"},
-      {:ex_aws_sns, "~> 2.2.0"},
+      # TOOD: use official release when tagged
```
arkadyan, line 44: suggestion (non-blocking): It might be worth calling out explicitly that the `check_if_phone_number_is_opted_out` functionality is what we are specifically wanting in the release (so that future us knows whether or not an acceptable release was indeed tagged).

apps/alert_processor/lib/dissemination/sms_opt_out_worker.ex
```diff
@@ -37,55 +39,76 @@ defmodule AlertProcessor.SmsOptOutWorker do
...
     end
   end
 
-  @spec list_numbers_opted_out_query(next_token) :: ExAws.Operation.Query.t()
-  defp list_numbers_opted_out_query(nil), do: ExAws.SNS.list_phone_numbers_opted_out()
-  defp list_numbers_opted_out_query(token), do: ExAws.SNS.list_phone_numbers_opted_out(token)
+  defp collect_opted_out(phone_numbers) do
+    Enum.reduce(phone_numbers, {[], 0, 0}, fn number, {opted_out, untouched_count, error_count} ->
+      case check_opted_out(number) do
+        {:ok, is_opted_out} ->
```
arkadyan, line 81: suggestion (non-blocking): You could also break out this case into

    {:ok, true} ->

and

    {:ok, _} ->

which would eliminate the nested `if-else`.

arkadyan (approved):
A couple of tiny comments to take or leave, but this looks great. Not failing the entire run on an error is a huge improvement.

PR #1235, opened 2022-07-26
Add rate limit to `SMSOptOutWorker`

zovt:
Running into some throttling according to Splunk, so this PR adds some rate limiting via sleep

### DIFF
--- a/apps/alert_processor/lib/dissemination/sms_opt_out_worker.ex
+++ b/apps/alert_processor/lib/dissemination/sms_opt_out_worker.ex
@@ -77,21 +77,30 @@ defmodule AlertProcessor.SmsOptOutWorker do
 
   defp collect_opted_out(phone_numbers) do
     Enum.reduce(phone_numbers, {[], 0, 0}, fn number, {opted_out, untouched_count, error_count} ->
-      case check_opted_out(number) do
-        {:ok, true} ->
-          {[number | opted_out], untouched_count, error_count}
+      res =
+        case check_opted_out(number) do
+          {:ok, true} ->
+            {[number | opted_out], untouched_count, error_count}
 
-        {:ok, _} ->
-          {opted_out, untouched_count + 1, error_count}
+          {:ok, _} ->
+            {opted_out, untouched_count + 1, error_count}
 
-        {:error, e} ->
-          Logger.warn(["SmsOptOutWorker event=aws_error #{inspect(e)}"])
-          {opted_out, untouched_count, error_count + 1}
-      end
+          {:error, e} ->
+            Logger.warn(["SmsOptOutWorker event=aws_error #{inspect(e)}"])
+            {opted_out, untouched_count, error_count + 1}
+        end
+
+      # From https://docs.aws.amazon.com/general/latest/gr/sns.html
+      # The rate limit for CheckIfPhoneNumberIsOptedOut is 50 requests /
+      # second. Sleep so that we're just under that limit.
+      Process.sleep(25)
+
+      res
     end)
   end
 
   defp do_process_opt_outs do
+    Logger.info("SmsOptOutWorker event=starting")
     phone_numbers = fetch_phone_numbers()
     {opted_out_numbers, untouched_count, error_count} = collect_opted_out(phone_numbers)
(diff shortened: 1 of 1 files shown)


arkadyan (commented):
I have to say, my initial reaction whenever I see `Process.sleep` is immediate fear. Just to warn you upfront that that's my starting point bias that's going to be coloring all of this feedback. 😉

The most practical concern I can think of is what happens if the processing time ends up being longer than the `opted_out_list_fetch_interval` configuration? Would that mean that we'd end up processing a second batch of phone numbers while the previous batch was still being processed? How big a problem would that be? I don't have admin access to easily see how many phone numbers we have in production, but I think that we'd hit that point at 144,000 which doesn't feel outrageous.

I tried to think through what an alternate approach might be. You could use some sort of big queue system, but since we can always start over with a DB query, that feels overkill. So maybe there's something better than this, but my initial thought was to split this out into 3 `GenServer` workers. Don't put much stock into my names, but my thinking was something like:

    SmsOptOutWorker --> OptOutCheckerWorker --> OptedOutProcessorWorker

- `SmsOptOutWorker`: Continues to work off the `opted_out_list_fetch_interval`, runs the DB query and tosses the results over to `OptOutCheckerWorker`
- `OptOutCheckerWorker`: Stores an internal queue of phone numbers, has its own interval configured to limit it to that 50 requests/second, calls `check_opted_out` if the number is opted out, throws it over to `OptedOutProcessorWorker`, and either way removes the number from its internal queue
- `OptedOutProcessorWorker`: Again has its own internal queue, has its own interval configured that I suspect could be very arbitrary—maybe on the longer side, calls `User.set_sms_opted_out\1`, logs the result, and that's about it

Sorry that this has been a bit long and rambly, but what do you think about this general idea? It's definitely added complexity, but maybe not too bad? I think each of the 3 workers would ideally be pretty simple, and the benefit is that it gives you the rate limiting for AWS while isolating it from affecting other things. But maybe it seems like overkill, or you can think of a better approach?

_As a completely tangential side note, I can't remember how much Elixir experience you have, but are you familiar with [GenStage](https://elixirschool.com/en/lessons/data_processing/genstage)? I don't think it's right for this use case—it's a "pull" model rather than push—but if it's new to you you might find it an interesting tool to know about._

zovt:
Hey @arkadyan, appreciate the feedback! I thought about this a bit and see where you're coming from. However, I think the lift might be a bit high considering the bandwidth I have left for this issue, and I'm not sure the juice is worth the squeeze.

It sounds like one of your primary concerns is around the `opted_out_list_fetch_interval` configuration, and potentially overlapping tasks. However, my understanding of [this code](https://github.com/mbta/alerts_concierge/blob/bdf756f3595e4b52a41995620730ac66e31fe80b/apps/alert_processor/lib/dissemination/sms_opt_out_worker.ex#L29-L34) is that we'll do a full run of `process_opt_outs` before queuing up another `:work` message to be sent an hour later, so there should never be a situation where we get an overlap (outside of the different environments but that's another issue...). Likewise, [this locking code](https://github.com/mbta/alerts_concierge/blob/bdf756f3595e4b52a41995620730ac66e31fe80b/apps/alert_processor/lib/dissemination/sms_opt_out_worker.ex#L39-L51) should act as a failsafe even if that weren't the case.

My understanding is that this worker is running in a separate elixir "process", so the `Process.sleep` call here shouldn't block the rest of the application. Is that correct? If not, then I think it could make sense to figure out some other solution which avoids this blocking.

Aside: the reason we even need this is because `ExAws`s automatic backoff/retry behavior doesn't recognize the rate limiting message format sent back by SNS. Could be worth opening an issue / fixing upstream at some point? (and notably, the backoff that `ExAws` applies relies on `sleep` as well 🙂)

Let me know if this makes sense and alleviates / aggravates the concerns you raised!

arkadyan (approved):
@zovt Yeah, I think everything you're saying seems correct and makes sense—you've convinced me. I also agree that fixing `ExAws` to properly recognize the rate-limiting message does indeed seem like the more proper improvement to consider. 👍 Thanks for talking it through with me.

PR #1238, opened 2022-08-10
feat: add some extra logging to further diagnose performance issues

zovt:
Let me know if I've missed anything or you'd like any additional information logged here


arkadyan:
@paulswartz As evidenced in the existing code in this file, those details are something I'm just learning now and good to know! 🙇

@zovt These extra bits of timing and including the alert ID seem good to me. The one extra tiny thing I might suggest considering is adding the alert ID to the "Scheduling distinct notifications" log line. I'm thinking it might be interesting to correlate the alert with the number of notifications that go out.

arkadyan (approved):
👍

paulswartz (approved):
🍰

PR #1239, opened 2022-08-19
New Terms and Conditions footer and checkbox

zovt:
Tickets:
- https://app.asana.com/0/1167196461144361/1202716358275341/f
- https://app.asana.com/0/1167196461144361/1202716358275340/f

Screenshots:

New footer text

New terms and conditions text and checkbox on the initial account creation options page

Example of error text when the checkbox isn't selected

New terms and conditions checkbox on account settings edit page


apps/concierge_site/lib/templates/account/edit.html.eex
```diff
@@ -35,6 +35,19 @@
           <%= label form, :phone_number, "My phone number is:", class: "form__label d-block" %>
           <%= telephone_input form, :phone_number, autocomplete: "off", placeholder: "###-###-####", class: "form-control d-inline-block form__phone--input", data: [toggle: "input"], required: false %>
           <%= error_tag form, :phone_number %>
+          <div class="mt-3 form-check">
```
arkadyan, line 38: suggestion(non-blocking): It looks like this block is repeated in 2 places, so it could get pulled out into a partial if you wanted.

arkadyan (approved):
👍

apps/concierge_site/lib/templates/account/edit.html.eex
zovt: yep, was thinking about that but wanted to keep it separate for now in case there were any design differences we wanted.

lund5000 (changes requested):
Minor copy change to the error message, otherwise this looks good to go: "You must consent to these terms to receive SMS alerts."

zovt:
@lund5000

lund5000 (approved):

PR #1240, opened 2022-09-29
fix: deploy to regular (non-`-a`) clusters

paulswartz:
As a part of migrating this infrastructure into Terraform, we're creating new clusters that don't have the `-a` suffix. This updates the main deploy step to deploy to the non `-a` clusters, but keeps around a deploy step to use them if we need it.

### DIFF
--- a/.github/workflows/deploy-a.yml
+++ b/.github/workflows/deploy-a.yml
@@ -1,10 +1,6 @@
-name: Deploy
+name: Deploy to -a (old)
 
 on:
-  workflow_run:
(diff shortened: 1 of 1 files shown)


zovt (approved):

arkadyan:
@paulswartz Do you happen to know if we are safe to remove the `deploy-a` action at this point?

paulswartz:
> @paulswartz Do you happen to know if we are safe to remove the `deploy-a` action at this point?

@arkadyan you are, yes!

PR #1244, opened 2022-11-09
build: Remove the old deploy-a action

arkadyan:
This has been replaced by the `deploy` action and the "-a" clusters are no longer in operation.

See https://github.com/mbta/alerts_concierge/pull/1240 for more details.


paulswartz (approved):
🍰

PR #1258, opened 2022-12-01
fix: Break apart database query with many ors

arkadyan:
Allow the pieces to use indexes. The old query had to run a sequential scan.

Even if the cumulative time doesn't decrease drastically, hopefully individual queries will be much shorter than the timeout we are currently sometimes seeing.

Before:

```
explain SELECT s0."id", s0."user_id", s0."trip_id", s0."relevant_days", s0."start_time", s0."end_time", s0."travel_start_time", s0."travel_end_time", s0."origin", s0."destination", s0."type", s0."route", s0."route_type", s0."direction_id", s0."origin_lat", s0."origin_long", s0."destination_lat", s0."destination_long", s0."rank", s0."return_trip", s0."facility_types", s0."paused", s0."parent_id", s0."is_admin", s0."inserted_at", s0."updated_at" FROM "subscriptions" AS s0 WHERE ( ( ( ( ( s0."is_admin" = true ) OR s0."route_type" = ANY ( '{0, 1, 2, 3, 4}' ) ) OR s0."route" = ANY ( '{Red}' ) ) OR s0."origin" = ANY ( '{}' ) ) OR s0."destination" = ANY ( '{}' ) ) AND ( s0."paused" = false ) AND ( NOT ( s0."user_id" = ANY ((SELECT DISTINCT ON (n.user_id) n.user_id FROM notifications AS n WHERE  n.alert_id = '133483' AND n.status = 'sent')) ) );
                                                                                                                       QUERY PLAN                                                                                                                        
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Seq Scan on subscriptions s0  (cost=400.24..14755.30 rows=114671 width=299)
   Filter: ((NOT paused) AND (NOT (hashed SubPlan 1)) AND (is_admin OR (route_type = ANY ('{0,1,2,3,4}'::integer[])) OR ((route)::text = ANY ('{Red}'::text[])) OR ((origin)::text = ANY ('{}'::text[])) OR ((destination)::text = ANY ('{}'::text[]))))
   SubPlan 1
     ->  Unique  (cost=378.97..393.34 rows=2758 width=16)
           ->  Sort  (cost=378.97..386.16 rows=2874 width=16)
                 Sort Key: n.user_id
                 ->  Index Scan using notifications_alert_id_index on notifications n  (cost=0.56..213.88 rows=2874 width=16)
                       Index Cond: ((alert_id)::text = '133483'::text)
                       Filter: ((status)::text = 'sent'::text)
```

After:

```
explain SELECT s0."id", s0."user_id", s0."trip_id", s0."relevant_days", s0."start_time", s0."end_time", s0."travel_start_time", s0."travel_end_time", s0."origin", s0."destination", s0."type", s0."route", s0."route_type", s0."direction_id", s0."origin_lat", s0."origin_long", s0."destination_lat", s0."destination_long", s0."rank", s0."return_trip", s0."facility_types", s0."paused", s0."parent_id", s0."is_admin", s0."inserted_at", s0."updated_at" FROM "subscriptions" AS s0 WHERE (s0."route" = ANY(array['Red'])) AND (s0."paused" = FALSE) AND (NOT (s0."user_id" = ANY((select distinct on (n.user_id) n.user_id from notifications as n where n.alert_id = '133483' and n.status = 'sent'))));
                                                          QUERY PLAN                                                          
------------------------------------------------------------------------------------------------------------------------------
 Bitmap Heap Scan on subscriptions s0  (cost=863.79..10877.05 rows=20664 width=299)
   Recheck Cond: (((route)::text = ANY ('{Red}'::text[])) AND (NOT paused))
   Filter: (NOT (hashed SubPlan 1))
   ->  Bitmap Index Scan on subscriptions_route_ix  (cost=0.00..458.38 rows=41328 width=0)
         Index Cond: ((route)::text = ANY ('{Red}'::text[]))
   SubPlan 1
     ->  Unique  (cost=378.97..393.34 rows=2758 width=16)
           ->  Sort  (cost=378.97..386.16 rows=2874 width=16)
                 Sort Key: n.user_id
                 ->  Index Scan using notifications_alert_id_index on notifications n  (cost=0.56..213.88 rows=2874 width=16)
                       Index Cond: ((alert_id)::text = '133483'::text)
                       Filter: ((status)::text = 'sent'::text)
```

### DIFF
--- a/apps/alert_processor/lib/model/subscription.ex
+++ b/apps/alert_processor/lib/model/subscription.ex
@@ -432,18 +432,22 @@ defmodule AlertProcessor.Model.Subscription do
 
   @spec all_active_for_alert(Alert.t()) :: [t()]
   def all_active_for_alert(alert) do
-    alert
-    |> get_alert_entity_lists()
-    |> entity_specific_queries(alert)
+    subscription_ids =
+      alert
+      |> get_alert_entity_lists()
+      |> entity_specific_queries(alert)
+      |> Enum.uniq()
+
+    fetch_with_user(subscription_ids)
   end
 
-  @spec entity_specific_queries(Keyword.t(), Alert.t()) :: [t()]
+  @spec entity_specific_queries(Keyword.t(), Alert.t()) :: [id()]
   defp entity_specific_queries(
          [route_ids: _, routes: _, stops: _, wildcard: true],
          alert
        ) do
-    from(s in __MODULE__)
-    |> shared_query(alert)
+    from(s in __MODULE__, select: s.id)
+    |> where_not_paused_and_not_yet_notified(alert)
   end
 
   defp entity_specific_queries(entity_lists, alert) do
@@ -457,9 +461,10 @@ defmodule AlertProcessor.Model.Subscription do
   defp admin_query(alert) do
     from(
       s in __MODULE__,
+      select: s.id,
       where: s.is_admin == true
     )
-    |> shared_query(alert)
+    |> where_not_paused_and_not_yet_notified(alert)
   end
 
   @spec entity_specific_query(tuple(), Alert.t()) :: [t()]
@@ -468,9 +473,10 @@ defmodule AlertProcessor.Model.Subscription do
   defp entity_specific_query({:route_ids, route_ids}, alert) do
     from(
       s in __MODULE__,
+      select: s.id,
       where: s.route_type in ^route_ids
     )
-    |> shared_query(alert)
+    |> where_not_paused_and_not_yet_notified(alert)
   end
 
   defp entity_specific_query({:routes, []}, _alert), do: []
@@ -478,9 +484,10 @@ defmodule AlertProcessor.Model.Subscription do
   defp entity_specific_query({:routes, routes}, alert) do
     from(
       s in __MODULE__,
+      select: s.id,
       where: s.route in ^routes
     )
-    |> shared_query(alert)
+    |> where_not_paused_and_not_yet_notified(alert)
   end
 
   defp entity_specific_query({:stops, []}, _alert), do: []
@@ -488,17 +495,17 @@ defmodule AlertProcessor.Model.Subscription do
   defp entity_specific_query({:stops, stops}, alert) do
     from(
       s in __MODULE__,
+      select: s.id,
       where: s.origin in ^stops or s.destination in ^stops
     )
-    |> shared_query(alert)
+    |> where_not_paused_and_not_yet_notified(alert)
   end
 
-  defp shared_query(query, alert) do
+  defp where_not_paused_and_not_yet_notified(query, alert) do
     query
     |> where_subscription_not_paused()
     |> where_not_yet_notified(alert)
     |> Repo.all()
-    |> Repo.preload(:user)
   end
 
   defp where_subscription_not_paused(query), do: from(s in query, where: s.paused == false)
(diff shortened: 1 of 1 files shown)


arkadyan:
@paulswartz Let me know if your plate is full, but tagging you initially since you were involved in these database investigations back with Nick. I believe this is roughly the next thing we'd discussed trying. Let me know if you remember differently or have other ideas to suggest.

Unfortunately, dev-green's database is empty enough that it's hard to get realistic timing. We can try deploying to production though and use the query time graphs on the [Splunk dashboard](https://mbta.splunkcloud.com/en-US/app/search/talerts_production?form.env=prod&form.npt-time.earliest=-24h%40h&form.npt-time.latest=now&form.email_time.earliest=-24h%40h&form.email_time.latest=now&form.nv-time.earliest=-24h%40h&form.nv-time.latest=now&form.nv-field=result&form.latency-time.earliest=-24h%40h&form.latency-time.latest=now&hideFilters=false) to diagnose what improvement this makes, if any.

apps/alert_processor/lib/model/subscription.ex
```diff
@@ -434,36 +434,96 @@ defmodule AlertProcessor.Model.Subscription do
...
+
+  defp shared_query(query, alert) do
+    query
+    |> where_subscription_not_paused()
+    |> where_not_yet_notified(alert)
+    |> Repo.all()
+    |> Repo.preload(:user)
+  end
+
+  # defp subscribers_match_query(
```
paulswartz, line 504: nitpick: you don't need to keep commented-out code

apps/alert_processor/lib/model/subscription.ex
```diff
@@ -434,36 +434,96 @@ defmodule AlertProcessor.Model.Subscription do
...
+      where: s.origin in ^stops or s.destination in ^stops
+    )
+    |> shared_query(alert)
+  end
+
+  defp shared_query(query, alert) do
+    query
+    |> where_subscription_not_paused()
+    |> where_not_yet_notified(alert)
+    |> Repo.all()
```
paulswartz, line 500: suggestion(non-blocking): you might consider splitting this part up, so that we only fetch the user once even if they're present in multiple alerts. a way to do this would be to have the "entity specific queries" and "admin_query" return a list of IDs, rather than a full subscription. then at the end, perform one query by the subscription ID, and preload the user on that single query.

paulswartz (approved):
:cake:
one non-blocking suggestion and one nitpick, otherwise it looks good!

apps/alert_processor/lib/model/subscription.ex
arkadyan: Oops! Thanks.

apps/alert_processor/lib/model/subscription.ex
arkadyan: Thanks, great idea. That made me realize I could through a `uniq` constraint in there while I was at it. Since this is a non-trivial change, would you mind giving it one last quick once-over, please?

paulswartz (approved):
🍰

apps/alert_processor/lib/model/subscription.ex
arkadyan: And actually wasn't even being used! So this saved it from deletion.

PR #1260, opened 2022-12-05
feat: Log timing for individual "entity" queries

arkadyan:
Asana ticket: [T-Alerts Notification Downtime 2022-07-09 Investigation](https://app.asana.com/0/1167196461144361/1202578832024792)

Follow-up to [#1258](https://github.com/mbta/alerts_concierge/pull/1258).

I realized that in addition to timing the entire group of queries that make up the existing "all active for alert" log message, we should also time each individual sub-query.

I will add an graph for each of these to the [Splunk dashboard](https://mbta.splunkcloud.com/en-US/app/search/talerts_production?form.env=prod&form.npt-time.earliest=-24h%40h&form.npt-time.latest=now&form.email_time.earliest=-24h%40h&form.email_time.latest=now&form.nv-time.earliest=-24h%40h&form.nv-time.latest=now&form.nv-field=result&form.latency-time.earliest=-24h%40h&form.latency-time.latest=now&hideFilters=false).

### DIFF
--- a/apps/alert_processor/lib/model/subscription.ex
+++ b/apps/alert_processor/lib/model/subscription.ex
@@ -3,6 +3,8 @@ defmodule AlertProcessor.Model.Subscription do
   Set of criteria on which a user wants to be sent alerts.
   """
 
+  require Logger
+
   alias AlertProcessor.{Repo, TimeFrameComparison}
 
   alias AlertProcessor.Model.{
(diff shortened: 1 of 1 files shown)


apps/alert_processor/lib/model/subscription.ex
```diff
@@ -446,8 +448,16 @@ defmodule AlertProcessor.Model.Subscription do
...
-    from(s in __MODULE__, select: s.id)
-    |> where_not_paused_and_not_yet_notified(alert)
+    start_time = System.monotonic_time(:millisecond)
+
+    subscription_ids =
+      from(s in __MODULE__, select: s.id)
+      |> where_not_paused_and_not_yet_notified(alert)
+
+    end_time = System.monotonic_time(:millisecond)
+    Logger.info("wildcard query, time=#{end_time - start_time}")
```
paulswartz, line 458: suggestion(non-blocking): this looks like something you could wrap up into a function (easier) or a macro (more efficient)
```elixir
def log_timing(key, fun) when is_function(fun, 0) do
  start_time = System.monotonic_time(:millisecond)
  result = fun.()
  end_time = System.monotonic_time(:millisecond)
  Logger.info("#{key} time=#{end_time - start_time}")
  result
end

log_timing("wildcard query", fn ->
  from(s in __MODULE__, select: s.id)
  |> where_not_paused_and_not_yet_notified(alert)
end)

paulswartz (approved):
one non-blocking suggestion, otherwise 🍰

PR #1282, opened 2023-02-08
feat: Use the built-in UUI generate function

arkadyan:
Part of getting [Set up a production-like environment to support load testing](https://app.asana.com/0/1167196461144361/1203535165046759) working.

The `gen_random_uuid` function is provided by Postgres starting in
version 13.

Remove the uuid-ossp extension that is no longer needed.

Also:
fix: Disable migration lock on concurrent migration
This was raised as necessary in the logs while running a migration.


paulswartz (approved):
The code looks good. The test failure looks like it's because it's referencing route 710, which no longer exists: https://www.mbta.com/schedules/710

arkadyan:
@paulswartz Thanks. Took me a bit to figure out because I had cached data locally.

PR #1295, opened 2023-02-23
fix (test): Adjust test expectations to match latest GTFS data

arkadyan:
No ticket. Recent GTFS changes broke a couple of tests.


KaylaBrady (approved):
👍 Not knowing much about these tests I'm curious - is it beneficial that they are tied to the current GTFS files, or is there an intention to decouple them down the line?

arkadyan:
@KaylaBrady Good question—it is indeed mostly just a pain. We do have a ticket to decouple them, but being a low-grade annoyance it's made it easy for it not to raise to the top of the list yet.

PR #1296, opened 2023-02-27
fix: Remove the DROP EXTENSION command that's failing

arkadyan:
The application DB user didn't have permission to run this command, and deploys have been failing silently since this was introduced.

We'll have to figure out another way to achieve removing the extension.


arkadyan:
cc @ianwestcott: Unfortunately this wasn't the fix I initially thought it was.

paulswartz (commented):
Is there a way to have the migration check whether it has the permissions to do the drop?

arkadyan:
@paulswartz That's an interesting question. A super quick search didn't turn anything up, but I'll poke around harder. 

It also occurred to me that removing the extension probably isn't very important in the end anyway. The original reason for moving away from it was that we had trouble installing it in a new DB instance. But even if nothing is using it, having the extension hanging out vestigially on some DB instances probably isn't actually hurting anything.

paulswartz:
> But even if nothing is using it, having the extension hanging out vestigially on some DB instances probably isn't actually hurting anything.
@arkadyan That was my thought as well, as long as it's not creating any functions that we could use accidentally.

ianwestcott:
I'd also be okay with a post-merge task for an infra engineer to remove the extension manually (first for dev where we can test it, then for prod). @arkadyan if you want to go that route, can you create an infrastructure request for it?

arkadyan:
This ticket got a little lost in the shuffle for me. But I like that idea, @ianwestcott, thanks.

@paulswartz I think that makes this ticket definitely good to good if it looks ok to you.

paulswartz (approved):
🍰

PR #1301, opened 2023-03-07
deps: Update ex_aws_sns to the latest standard release

arkadyan:
No Asana ticket, I just happened to find the note in the code and could address it.

Support for `check_if_phone_number_is_opted_out` that we needed has now been released.


KaylaBrady (approved):
Nice!

PR #1302, opened 2023-03-07
fix: Notifications are built with proper fields for comms mode

arkadyan:
Asana task: [Go through T-Alerts and make sure notification choice is being used everywhere](https://app.asana.com/0/1167196461144361/1204118979786026)

In the world of the notification struct, the existence of the phone number or email field indicates how to send the message. Make sure that we only populate one or the other, depending on what the user has selected as their preferred communication mode.


apps/alert_processor/test/alert_processor/dissemination/mass_notifier_test.exs
```diff
@@ -62,10 +62,23 @@ defmodule AlertProcessor.Dissemination.MassNotifierTest do
...
-      {_, second_phone} = insert_user_and_build_notification(%{phone_number: "5555556789"})
-      {_, second_email} = insert_user_and_build_notification(%{phone_number: nil})
+      {_, first_phone} =
+        insert_user_and_build_notification(%{
+          communication_mode: "sms",
+          phone_number: "5555551234"
+        })
+
+      {_, first_email} =
+        insert_user_and_build_notification(%{communication_mode: "email", phone_number: nil})
```
KaylaBrady, line 72: suggestion (non-blocking): It was a bit confusing to me as someone not familiar with this codebase why phone number was being set explicitly but email wasn't. I see now that it is set by default from the user factory, so no change really needed here. Maybe for consistency since the phone_number is explicitly set for the `first_phone` and `second_phone` users, consider setting the email for `first_email` and `second_email` users too.

apps/alert_processor/test/alert_processor/rules_engine/notification_builder_test.exs
```diff
@@ -6,45 +6,77 @@ defmodule AlertProcessor.NotificationBuilderTest do
...
-          subscription_id: sub.id
-        }
-      ],
-      closed_timestamp: alert.closed_timestamp,
-      type: :initial
-    }
-
-    assert expected_notification == notification
+  describe "build_notification" do
+    setup do
```
KaylaBrady, line 10: 🙌 Great to see this broken out into setup!

apps/alert_processor/test/alert_processor/rules_engine/subscription_filter_engine_test.exs
```diff
@@ -40,8 +40,8 @@ defmodule AlertProcessor.SubscriptionFilterEngineTest do
     test "the expected subscription matches", %{alert: alert} do
       email_match = "route1,route4|weekdays|low,high|10am-2pm@test.com"
       email_no_match = "route2,route3|weekdays,sunday|low,high|10am-2pm@test.com"
-      user = insert(:user, phone_number: nil, email: email_match)
-      user2 = insert(:user, phone_number: nil, email: email_no_match)
+      user = insert(:user, communication_mode: "email", phone_number: nil, email: email_match)
+      user2 = insert(:user, communication_mode: "email", phone_number: nil, email: email_no_match)
```
KaylaBrady, line 44: question (non-blocking): Was this test failing without this change? Or is this added so the test is more explicit? Trying to understand whether this default value of ["email" for communication_mode](https://github.com/mbta/alerts_concierge/blob/04ec1a7841f76ef578d911c3faa2d9276ab4b985/apps/alert_processor/lib/model/user.ex#L34) is as expected

KaylaBrady (approved):
Some small questions for my own curiosity, but looks good!

apps/alert_processor/test/alert_processor/rules_engine/subscription_filter_engine_test.exs
arkadyan: Good question. It was indeed working fine previously, adding it really was an attempt to make it more explicit. Do you find that helpful at all? If so, is it worth a comment or something clarifying that it's just intended to be explicit?

arkadyan: I suppose alternately we could remove this explicit `communication_mode: "email"` bit that's redundant, but instead add a comment.

KaylaBrady: I like that this is more explicit now! Just wasn't sure whether this was a functional change too, but good to know it is not!

PR #1309, opened 2023-05-08
feat: Support Keycloak auth

arkadyan:
Asana task: [Add Keycloak auth support](https://app.asana.com/0/1167196461144361/1203867762984593)

Support auth via either local database (what we've always used up to now) or Keycloak (the new hotness), configured via an environment variable.

Once we've cut all environments over to using Keycloak and migrated all of our production users, the plan is to eventually remove local auth and only support Keycloak.


arkadyan:
@lemald Apologies for the size of this. I tried to break it up by logical pieces as much as possible. I'd be happy to walk through it over video if that would be helpful.

apps/alert_processor/lib/model/user.ex
```diff
@@ -324,6 +324,9 @@ defmodule AlertProcessor.Model.User do
     |> Repo.transaction()
   end
 
+  @spec get(id()) :: t() | nil
+  def get(id), do: Repo.get_by(__MODULE__, id: id)
```
lemald, line 328: **nitpick (non-blocking):** any reason to use `get_by` rather than just `get` if you're fetching it by the `id`?

arkadyan: Restating the questions to confirm that I'm understanding correctly: you are suggesting that in a controller (as a random example), instead of calling `User.get(id)`, I could instead just call `Repo.get_by(User, id: 42)`, is that right?

The honest answer is that I don't think I considered that, so it's an interesting question. Thinking about it now retroactively, there are 3 benefits I can think of for the approach I took:

1. It encapsulates "user stuff" a little more into the domain of the User module
2. Callers don't need to know about `Repo`
3. The syntax is simpler

Are there pros to the other approach I'm not considering?

Alternately, now that I'm looking at at I'm noticing that we currently have the functions `User.get/1` and `User.for_email/1`. Maybe those should be integrated better so maybe you have something like `User.by(id: 42)` and `User.by(email: "jdoe@example.com")`?

arkadyan: Oh, hey! Yeah, that's better, I forgot about that one.

apps/concierge_site/lib/controllers/auth_controller.ex
```diff
@@ -0,0 +1,105 @@
...
+  alias ConciergeSite.SessionHelper
+  alias Plug.Conn
+
+  plug(Ueberauth)
+
+  @spec register(Conn.t(), map()) :: Conn.t()
+  def register(conn, _params) do
+    base_uri = System.get_env("KEYCLOAK_BASE_URI")
+    client_id = System.get_env("KEYCLOAK_CLIENT_ID")
+    redirect_uri = "KEYCLOAK_REDIRECT_URI" |> System.get_env() |> URI.encode()
```
lemald, line 19: **suggestion (non-blocking):** it might be more Elixir-y to load these into application environment variables via `config/runtime.exs`.

apps/concierge_site/lib/controllers/auth_controller.ex
```diff
@@ -0,0 +1,105 @@
...
+
+    registration_uri =
+      "#{base_uri}/auth/realms/MBTA/protocol/openid-connect/registrations?client_id=#{client_id}&response_type=code&scope=openid&redirect_uri=#{redirect_uri}"
```
lemald, line 22: **question:** any possible issues here related to not URI-encoding the values, especially `redirect_uri`?

apps/concierge_site/lib/controllers/auth_controller.ex
```diff
@@ -0,0 +1,105 @@
...
+      ) do
+    phone_number =
+      user_info
+      |> Map.get("phone")
+      |> strip_us_country_code()
+
+    user =
+      id
+      |> get_or_create_user(email, phone_number)
+      |> use_email_and_phone_from_token(email, phone_number)
```
lemald: **question:** just so I understand the transition plan correctly - it looks like you're using the email and phone number from the SSO response to create a new user if an old user isn't present, but otherwise you don't update those fields in the DB... though I'm guessing that since you're looking up by the ID in the SSO response this would _only_ give you new users? Or is the MBTA SSO ID specially created to bridge the old and new accounts?

apps/concierge_site/lib/helpers/session_helper.ex
```diff
@@ -0,0 +1,52 @@
...
+    |> redirect(to: sign_in_redirect_path(user))
+  end
+
+  @spec sign_out(Conn.t()) :: Conn.t()
+  def sign_out(conn) do
+    redirect_to =
+      if keycloak_auth?() do
+        [
+          external:
+            "#{System.get_env("KEYCLOAK_LOGOUT_URI")}?redirect_uri=#{page_url(conn, :landing)}"
```
lemald, line 26: **question:** see above question about URI encoding.

apps/concierge_site/test/web/controllers/auth_controller_test.exs
```diff
@@ -67,6 +71,7 @@ defmodule ConciergeSite.Web.AuthControllerTest do
          %{conn: conn, rider: %User{id: id}} do
       new_email = "newemail@example.com"
       new_phone_number = "+15559876543"
+      new_role = "admin"
```
lemald, line 74: **suggestion:** it might be nice to have an explicit case for the "user" role case, since the failure where that isn't handled correctly and results in improperly assigning an "admin" role would be bad.

lemald:
@arkadyan okay, sorry for the delay but I've finally managed to make it through a full pass of review! I have a few comments and questions but overall this is looking good.

apps/concierge_site/lib/controllers/auth_controller.ex
arkadyan: Yeah, good questions, and maybe they imply that I should add some comments or something to describe this better.

I'll start with the last part: the plan is to export all existing user accounts into Keycloak, which will include the existing IDs. So lookups should cover all existing users, whether they created their account pre- or post-Keycloak transition. The case that won't be found in the database is a user who just created their account in Keycloak and is arriving here for the first time. So those are the ones we need to write to the database.

Getting to the bigger question, we have a system where the user changes their account details (email, phone) in Keycloak, but since we need to save that data locally to be able to send notifications, Keycloak sends a message for each of those updates via SQS. T-Alerts will regularly check SQS for new messages, and apply the given change to the local database so that the local data mirrors the authoritative source in Keycloak. But there's a possible/likely scenario where a user goes to Keycloak to edit their account, and then bounces back to T-Alerts. So the data in the token reflects the latest updates, but it's likely that we haven't yet gotten the message and updated our own database, so our local data might show the old values. So we want to use the new values they just set for what we show on the page since those are accurate. But since we know that the SQS system will eventually update the values in or DB to match, we don't need to actually worry about writing to the database here.

Sorry, that's long. I'll try to put something more concise into a comment.

apps/concierge_site/lib/controllers/auth_controller.ex
arkadyan: Sure seems like no reason not to.

lemald:
@arkadyan okay, looks like the only things remaining are my comment about that additional test case and the failing test in CI, once those are addressed I think I'll be ready to approve. 👍

apps/concierge_site/test/web/controllers/auth_controller_test.exs
arkadyan: Yep, that's a very good idea. Added a test for that case.

lemald (approved):
🎉

config/runtime.exs
```diff
@@ -3,9 +3,30 @@ import Config
 config :alert_processor, AlertProcessor.Repo,
   url: System.fetch_env!("DATABASE_URL_#{config_env() |> to_string() |> String.upcase()}")
 
+config :concierge_site, ConciergeSite.Endpoint,
+  authentication_source: System.get_env("AUTHENTICATION_SOURCE", "local")
+
+config :concierge_site,
+  keyclooak_base_uri: System.get_env("KEYCLOAK_BASE_URI"),
+  keyclooak_client_id: System.get_env("KEYCLOAK_CLIENT_ID"),
+  keyclooak_redirect_uri: System.get_env("KEYCLOAK_REDIRECT_URI")
```
firestack, line 12: Typo?

```suggestion
  keycloak_base_uri: System.get_env("KEYCLOAK_BASE_URI"),
  keycloak_client_id: System.get_env("KEYCLOAK_CLIENT_ID"),
  keycloak_redirect_uri: System.get_env("KEYCLOAK_REDIRECT_URI")
```

arkadyan: Ha ha, yep indeed. Thanks.

config/dev.exs
```diff
@@ -0,0 +1,6 @@
+import Config
+
+config :ueberauth, Ueberauth,
+  providers: [
+    keycloak: {Draft.Ueberauth.Strategy.FakeKeycloak, []}
```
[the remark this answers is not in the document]
arkadyan: It's the [old application](https://github.com/mbta/draft) I copy-pasted the initial setup from. You are catching everything! (Thank you)

config/runtime.exs
```diff
@@ -3,9 +3,30 @@ import Config
...
+
+config :ueberauth, Ueberauth.Strategy.OIDC,
+  keycloak: [
+    fetch_userinfo: true,
+    userinfo_uid_field: "preferred_username",
+    discovery_document_uri: System.get_env("KEYCLOAK_WELL_KNOWN_OIDC"),
+    client_id: System.get_env("KEYCLOAK_CLIENT_ID"),
+    client_secret: System.get_env("KEYCLOAK_CLIENT_SECRET"),
+    redirect_uri: System.get_env("KEYCLOAK_REDIRECT_URI"),
+    logout_uri: System.get_env("KEYCLOAK_LOGOUT_URI"),
```
firestack, line 22: note(non-blocking): @lemald and I were just talking about implementing this on mycharlie and that maybe this(`config :ueberauth, Ueberauth.Strategy.OIDC`) isn't the best place for this config item(`logout_uri`) since Ueberauth doesn't handle the logout flow? (@lemald knows more about this than I)

If you're looking for ideas on where else to put it, I think I'm going to play with the idea of adding `logout_uri` config to `MyCharlieWeb.AuthController` on our side.

arkadyan: Yeah, that's a fair point. It's kind of funky because it relates to Keycloak, but as you say, not actually to Ueberauth.

arkadyan: Another note: I think I only learned about this in-depth after implementing this logout URI var, but Keycloak publishes a list of ["well-known" endpoints](https://login-sandbox.mbtace.com/auth/realms/MBTA/.well-known/openid-configuration). (It seems like that's probably part of the OIDC standard, since we pass that down to the [openid_connect](https://github.com/DockYard/openid_connect) library.) And that list does include that logout URL under `end_session_endpoint`.  It would be more effort, so maybe not worth it, but maybe the more "proper" thing would be to query that well-known list and discover the logout URL instead of saving it. (But all that said, I'm probably not going to bother with that right this minute…)

firestack: Oh interesting, This'll be helpful for me exploring logout on our side, thanks!!

arkadyan: Poking around a little, maybe you could even [get it from openid_connect](https://github.com/DockYard/openid_connect/blob/ddafedc3c81a5bc91919d13630b6cb5ea2595ffe/lib/openid_connect.ex#L152)? That library is wrapped by [ueberauth_oidc](https://github.com/DefactoSoftware/ueberauth_oidc), but I did end up needing to use it directly in [one place](https://github.com/mbta/alerts_concierge/pull/1309/files#diff-bb7878acb7873ea7a1866a588482aee73b3b4ee08c2fbeb3e4e6b496b88d3c5dR121).

firestack: Heh, looks like the library was rewritten like a month ago, and that function seems to be new? https://github.com/DockYard/openid_connect/pull/50

arkadyan: Oh, interesting. And while I was working I was constantly reference the source files loaded locally, so that makes sense, I probably looked around it it wasn't there.

apps/concierge_site/lib/helpers/session_helper.ex
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  @spec sign_out(Conn.t()) :: Conn.t()
+  def sign_out(conn) do
+    redirect_to =
+      if keycloak_auth?() do
+        [
+          external:
+            URI.encode(
+              "#{System.get_env("KEYCLOAK_LOGOUT_URI")}?redirect_uri=#{page_url(conn, :landing)}"
```
firestack, line 27: question(non-blocking): Just wondering if this works for you locally? when I try to use the logout URI from 1p I get invalid `redirect_uri` but I don't know if I'm missing extra params or should have it be a specific url, I'm currently just sending it to `https://alerts.localhost:4005/` which returns a page saying "Invalid redirect uri"

arkadyan: @firestack Ah, yes, that's not obvious. First I set up `alerts.localhost` pointing to `127.0.0.1` in my `/etc/hosts` file, but it sounds like you've probably already passed that step. The other thing I had to do was configure that as an allowed redirect path in Keycloak. Are you pointing at the sandbox Keycloak? Maybe it's easier to chat in Slack to diagnose further.

arkadyan:
@lemald and @firestack, FYI I just push https://github.com/mbta/alerts_concierge/pull/1309/commits/6a84ea1b747d5f583422e4d7973a966e28cce65e that includes some updates to support a "register a new user account" action in addition to the standard login action. (Though much of the actual implementation details are in the [library](https://github.com/DefactoSoftware/ueberauth_oidc/compare/master...arkadyan:ueberauth_oidc:custom-uri).) I could walk through the details around this with you if you are interested in hearing more.

firestack:
> https://github.com/mbta/alerts_concierge/pull/1309#issuecomment-1589972216

Thanks @arkadyan!
I'm off MyCharlie now, but https://github.com/mbta/my_charlie/pull/29 is still open and those changes should also probably be integrated into the PR. 

CC'ing current my-charlie devs: @grejdi-mbta @napilla @panentheos

nihonjinrxs:
> FYI, I just pushed https://github.com/mbta/alerts_concierge/commit/6a84ea1b747d5f583422e4d7973a966e28cce65e that includes some updates to support a "register a new user account" action in addition to the standard login action. (Though much of the actual implementation details are in the [library](https://github.com/DefactoSoftware/ueberauth_oidc/compare/master...arkadyan:ueberauth_oidc:custom-uri).) I could walk through the details around this with you if you are interested in hearing more.

Hey, @arkadyan, I'm picking up this work now on MyCharlie and would love to walk through this with you sometime early next week if you have time. Thanks.

nihonjinrxs:
Also, in reading up on this, I noticed [this `ueberauth_keycloak_strategy` lib](https://github.com/Rukenshia/ueberauth_keycloak) listed in [the Ueberauth List of Strategies wiki page](https://github.com/ueberauth/ueberauth/wiki/List-of-Strategies) as in "Beta" status. Did we look at this before and decide against it, or is this new enough that it wasn't there before? Worth looking at it and potentially updating it to meet our needs?

PR #1314, opened 2023-05-10
feat: Survey a user who looks like they are deleting their account

arkadyan:
Asana ticket: [Add inline survey when riders delete account or all subscriptions](https://app.asana.com/0/1167196461144361/1204501447170890/f)

Show an Informizely survey if

- they actually delete their account
- they delete all of their subscriptions

Populate Informizely site and survey IDs from env vars since they need to be different values in dev vs. production.

### DIFF
--- a/apps/concierge_site/config/config.exs
+++ b/apps/concierge_site/config/config.exs
@@ -62,14 +62,6 @@ config :concierge_site, mailchimp_api_key: {:system, "MAILCHIMP_API_KEY", ""}
 config :concierge_site, mailchimp_list_id: {:system, "MAILCHIMP_LIST_ID", "abc123"}
 config :concierge_site, mailchimp_api_client: ConciergeSite.Mailchimp.FakeClient
 
-# Informizely
-config :concierge_site,
-  informizely_site_id: {:system, "INFORMIZELY_SITE_ID", "69f607f3-3826-4aa6-9f86-996a07670833"}
-
-config :concierge_site,
-  informizely_account_deleted_survey_id:
-    {:system, "INFORMIZELY_ACCOUNT_DELETED_SURVEY_ID", "rhluhvlu"}
-
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
 import_config "#{config_env()}.exs"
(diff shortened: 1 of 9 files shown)


apps/concierge_site/config/config.exs
```diff
@@ -62,6 +62,14 @@ config :concierge_site, mailchimp_api_key: {:system, "MAILCHIMP_API_KEY", ""}
...
 config :concierge_site, mailchimp_api_client: ConciergeSite.Mailchimp.FakeClient
 
+# Informizely
+config :concierge_site,
+  informizely_site_id: {:system, "INFORMIZELY_SITE_ID", "69f607f3-3826-4aa6-9f86-996a07670833"}
+
+config :concierge_site,
+  informizely_account_deleted_survey_id:
+    {:system, "INFORMIZELY_ACCOUNT_DELETED_SURVEY_ID", "rhluhvlu"}
+
```
KaylaBrady, line 72: question (non-blocking): would it be more appropriate for these to live in [config/runtime.exs](https://github.com/mbta/alerts_concierge/blob/main/config/runtime.exs)?

Are the default values checked in to support local testing?

apps/concierge_site/lib/templates/trip/index.html.eex
```diff
@@ -16,6 +16,11 @@
       <%= link(to: trip_path(@conn, :new), class: "btn btn-primary btn-block trip__card--btn") do %>
         <span>Create another subscription <i aria-hidden="true" class="fa fa-plus-circle"></i></span>
       <% end %>
+
+
+      <%= if show_deleted_last_trip_survey?(@conn, @trips) do %>
+        <%= informizely_account_deleted_survey() %>
+      <% end %>
```
KaylaBrady, line 23: question: Anything worth testing about this logic?

KaylaBrady (approved):
Maybe worth some tests, but looks good to me!

apps/concierge_site/config/config.exs
arkadyan: Yeah, that's better, moved them. I also removed the default values. I guess that was the reason, but it doesn't really seem necessary.

apps/concierge_site/lib/templates/trip/index.html.eex
arkadyan: Good idea, added some tests.

PR #1322, opened 2023-06-05
fix (test): Update test data to match current schedule

arkadyan:
No Asana ticket, fixes tests that are currently breaking in `main`.


arkadyan:
@ErinLMoore I pushed a fix with the trip IDs and times for the correct direction.

PR #1324, opened 2023-06-08
Time Window and Fully Paused Account Queries

ErinLMoore:
Screenshot 2023-06-07 at 4 12 24 PM
Screenshot 2023-06-08 at 8 59 31 AM

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204392487059360
  - https://app.asana.com/0/0/1204392487059364

### DIFF
--- a/apps/alert_processor/lib/model/query.ex
+++ b/apps/alert_processor/lib/model/query.ex
@@ -183,6 +183,14 @@ defmodule AlertProcessor.Model.Query do
         LEFT JOIN users_before_end ON users_before_end.period_end = periods.end
         ORDER BY periods.end DESC
         """
+      },
+      %__MODULE__{
+        label: "Grouped Time Windows",
+        query: """
+        SELECT s.start_time, s.end_time, s.type, count(distinct s.id)
+        FROM subscriptions s
+        GROUP BY s.type, s.start_time, s.end_time;
+        """
       }
     ]
     |> Enum.map(&%{&1 | id: &1.label |> String.downcase() |> String.replace(" ", "_")})
--- a/apps/alert_processor/lib/model/query.ex
+++ b/apps/alert_processor/lib/model/query.ex
@@ -187,9 +187,52 @@ defmodule AlertProcessor.Model.Query do
       %__MODULE__{
         label: "Grouped Time Windows",
         query: """
-        SELECT s.start_time, s.end_time, s.type, count(distinct s.id)
+        SELECT
+          to_char(s.start_time, 'FMHH24:MI') as start_time,
+          to_char(s.end_time, 'FMHH24:MI') as end_time, s.type,
+          count(distinct s.id)
         FROM subscriptions s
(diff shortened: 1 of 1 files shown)


apps/alert_processor/lib/model/query.ex
```diff
@@ -183,6 +183,57 @@ defmodule AlertProcessor.Model.Query do
...
+          to_char(s.end_time, 'FMHH24:MI') as end_time, s.type,
+          count(distinct s.id)
+        FROM subscriptions s
+        GROUP BY s.type, s.start_time, s.end_time;
+        """
+      },
+      %__MODULE__{
+        label: "Fully Paused Accounts",
+        query: """
+          WITH
```
ErinLMoore, line 201: Please let me know if there's some other obvious way to do this one.

arkadyan:
@ErinLMoore Quick note before I look at the details: could you please link to the Asana ticket[s] (if there is one) in the description of PRs? That's a general convention we follow because it's helpful to be able to refer back to the description of the work you are completing. (We also have a hook set up on this project where it will update the Asana tickets with a reference to the PR, which is handy for referencing the solution from the ticket.)

apps/alert_processor/lib/model/query.ex
```diff
@@ -183,6 +183,14 @@ defmodule AlertProcessor.Model.Query do
         LEFT JOIN users_before_end ON users_before_end.period_end = periods.end
         ORDER BY periods.end DESC
         """
+      },
+      %__MODULE__{
```
arkadyan, line 187: Cool, that's actually surprisingly simple.

apps/alert_processor/lib/model/query.ex
```diff
@@ -187,10 +187,53 @@ defmodule AlertProcessor.Model.Query do
...
+        label: "Fully Paused Accounts",
+        query: """
+          WITH
+          paused_accounts AS (
+            SELECT u.id, MAX(s.updated_at) as newest_subscription FROM users u
+            INNER JOIN subscriptions s ON u.id = s.user_id
+            GROUP BY u.id
+            HAVING sum(case when s.paused = false then 1 else 0 end) = 0
```
arkadyan, line 206: This is a clever solution.

apps/alert_processor/lib/model/query.ex
arkadyan: Nice! With the caveat that I'm far from a SQL master, I'd say that this all seems reasonable to me. And more than that, it's really incredibly readable for what it is, which I appreciate.

arkadyan (approved):
👍 Awesome!

apps/alert_processor/lib/model/query.ex
ErinLMoore: Yeah I am also not an SQL master 😅 The unions feel a bit weird but if it works it works (I tested it with several accounts and subscriptions, seems solid enough). Thanks!

PR #1339, opened 2023-07-14
New parents subscriptions query

ErinLMoore:
Query to list number of users who have at least one unpaused subscription that have child subscriptions.

### DIFF
--- a/apps/alert_processor/lib/model/query.ex
+++ b/apps/alert_processor/lib/model/query.ex
@@ -236,9 +236,9 @@ defmodule AlertProcessor.Model.Query do
         """
       },
       %__MODULE__{
-        label: "Count of Active Users Having Subscriptions That Are Parents",
+        label: "Count of Users With Parallel Bus Route Subscriptions",
         query: """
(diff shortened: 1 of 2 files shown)


apps/alert_processor/lib/model/query.ex
```diff
@@ -234,6 +234,18 @@ defmodule AlertProcessor.Model.Query do
           UNION ALL
           SELECT 'Paused for 12 months or more',plmore.total_paused_accounts FROM paused_for_twelve_months_or_more plmore;
         """
+      },
+      %__MODULE__{
+        label: "Count of Active Users Having Subscriptions That Are Parents",
```
arkadyan, line 239: thought (non-blocking): I wonder if there's some wording that makes this label more generally understandable / less specific to the implementation details. Maybe something along the lines of, "parallel route subscriptions"?

arkadyan (approved):
👍

apps/alert_processor/lib/model/query.ex
arkadyan: Or even, "parallel bus route subscriptions" since this is only possible with bus routes?

PR #1350, opened 2023-08-24
feat: Remove local auth support

arkadyan:
Asana ticket: [Post-Keycloak migration tech debt](https://app.asana.com/0/1167196461144361/1205295947643673)

Now that we have successfully migrated all of our environments to Keycloak, we no longer need to support local authentication.

- feat: Remove all keycloak_auth? branching

- fix: Skip redirecting to OIDC logout on failed login
In this case, we know that they were never actually logged in.

- feat: Remove unused authentication_source env var
We'll follow this up with a Terraform update to remove this variable once this PR is deployed across all of our servers.
Don't run OIDC worker in test.

- feat: Remove local password reset functionality
Including controller, model, and related templates.

- feat: Drop the password_resets database table

- feat: Remove Admins controller and routes
User roles are now controlled in Keycloak.

- feat: Remove user passwords
This is now handled by Keycloak.

- feat: Remove user email validation
This is now handled by Keycloak. The email address is actually verified
before the user can make it to T-Alerts.

_NB: In draft for a few days while we confirm that T-Alerts + Keycloak is working well in production for real riders._


ErinLMoore:
I see that if you run this locally it connects to the Keycloak in the dev sandbox- is there any way to create a user locally without using a real email address?

ErinLMoore (approved):
💥 Looks good, I pulled it down and everything seems to work fine.

arkadyan:
> I see that if you run this locally it connects to the Keycloak in the dev sandbox- is there any way to create a user locally without using a real email address?

That's a good question. There's the `FakeKeycloak` strategy that's mostly set up for testing right now, so it's mocking out some of the interactions, but I think not all. We could probably do a little bit to improve that so that in dev you don't interact with Keycloak at all and are just automatically "logged in" using a fake account. The one downside there is that it would just give you a single account each time, so maybe we could build in a trap door or something to let you actually use Keycloak when you want to have access to different accounts.

ErinLMoore:
> > I see that if you run this locally it connects to the Keycloak in the dev sandbox- is there any way to create a user locally without using a real email address?
> 
> That's a good question. There's the `FakeKeycloak` strategy that's mostly set up for testing right now, so it's mocking out some of the interactions, but I think not all. We could probably do a little bit to improve that so that in dev you don't interact with Keycloak at all and are just automatically "logged in" using a fake account. The one downside there is that it would just give you a single account each time, so maybe we could build in a trap door or something to let you actually use Keycloak when you want to have access to different accounts.

For right now I say let's leave it like it is, I just wanted to ask in case there was some way to do it I was missing.

PR #1353, opened 2023-08-31
Adds image and alt text to emails

ErinLMoore:
Reference for the shape the image data comes in as: https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#message-translatedimage

Asana ticket:  https://app.asana.com/0/1167196461144361/1204624899415315/f

### DIFF
--- a/apps/alert_processor/test/alert_processor/rules_engine/notification_builder_test.exs
+++ b/apps/alert_processor/test/alert_processor/rules_engine/notification_builder_test.exs
@@ -17,7 +17,9 @@ defmodule AlertProcessor.NotificationBuilderTest do
         id: "1",
         header: nil,
         active_period: [%{start: two_days_from_now, end: three_days_from_now}],
-        last_push_notification: one_hour_ago
+        last_push_notification: one_hour_ago,
+        image_url: "http://example.com/cool_image.png",
+        image_alternative_text: "a cool image"
       }
 
       {:ok, alert: alert}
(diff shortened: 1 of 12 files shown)


apps/alert_processor/test/alert_processor/rules_engine/notification_builder_test.exs
```diff
@@ -46,7 +46,9 @@ defmodule AlertProcessor.NotificationBuilderTest do
           }
         ],
         closed_timestamp: alert.closed_timestamp,
-        type: :initial
+        type: :initial,
+        image_url: nil,
```
arkadyan, line 50: suggestion: Maybe worth a test case that includes values for these fields?

apps/concierge_site/lib/mail_templates/notification.txt.eex
```diff
@@ -11,6 +11,10 @@ The issue described below has ended:
 
 <%= notification.header %>
 
+<%= notification.image_url %>
```
arkadyan, line 14: question: If these values are nil, does it result in an email with a bunch of extra empty lines? If so it might be worth including `if` statements.

apps/concierge_site/test/lib/dissemination/notification_email_test.exs
```diff
@@ -47,6 +48,7 @@ defmodule ConciergeSite.Dissemination.NotificationEmailTest do
     assert body =~ "Red line delay"
     assert body =~ "Red line inbound from Alewife station closure"
     assert body =~ "There is a fire in at south station so it is closed"
+    assert body =~ "http://www.example.com/alert-info.png"
```
arkadyan, line 51: suggestion: Could you please also add a quick assertion to the html email test?

arkadyan (changes requested):
A couple really small suggestions, but otherwise looks good.

apps/concierge_site/lib/mail_templates/notification.txt.eex
ErinLMoore: Ah, I put the "if" in the MJML but forgot it here. It results in an email with a size 0 image element but that's still not ideal.

arkadyan (approved):
👍

PR #1360, opened 2023-09-15
Add sorting to active users query

ErinLMoore:
I was looking at this in prod and realized it would be much nicer to look at if it were sorted. No ticket.


arkadyan (approved):
Looks like it needs a rebase, but otherwise 👍

PR #1362, opened 2023-09-26
Bumps mjml version, fixes image width on MS Outlook

ErinLMoore:
There's a known issue where images won't be properly width constrained on some versions of MS Outlook app for Windows. This fixes it for some reason (Tested on actual outlook apps by both Steve and I).

I also bumped the MJML version just to make it up-to-date with the docs and to stay compatible with newer versions of apps. Doesn't seem to have made any visible difference.


arkadyan (approved):
👍

PR #1364, opened 2023-10-13
Fix: Migrate GitHub actions to v2 and assumed role

krisrjohnson21:
This PR migrates the deploy workflow to use v2 of `build-push-ecr` and `deploy-ecs` actions. 

There's a test failure happening in this pipeline, so I'd like to get team opinion here. I wouldn't have expected these changes to affect any test coverage.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -29,16 +29,14 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: mbta/actions/build-push-ecr@v1
+      - uses: mbta/actions/build-push-ecr@v2
         id: build-push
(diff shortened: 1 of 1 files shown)


paulswartz (approved):
I agree that your changes shouldn't have impacted the tests at all. My guess is it's something about active alerts or a new GTFS output?

arkadyan:
@krisrjohnson21 Test failures in this app due to GTFS releases is unfortunately a pretty regular occurrence. I can look into it on Monday.

arkadyan (approved):
👍 

@krisrjohnson21 I rebased this branch onto some text fixes. CI is running now but should pass I think.

PR #1366, opened 2023-11-06
Email accessibility improvements

lund5000:
Asana ticket: https://app.asana.com/0/1167196461144361/1205857212970322/f

NOTE: These changes have not been tested in the application yet. They have been tested using a combination of the [MJML Email Editor](https://mjml.io/try-it-live) and https://www.accessible-email.org/.


apps/concierge_site/assets/mjml/notification.mjml
```diff
@@ -99,10 +99,10 @@
     <mj-raw>
       <!-- Footer -->
     </mj-raw>
-    <mj-section padding="0">
+    <mj-section padding="0" css-class="footer">
       <mj-column>
         <mj-text font-size="12px" font-family="'Helvetica Neue', Helvetica, Arial, sans-serif" padding-left="0" padding-right="0">
-          <p>
+          <p role="contentinfo">
```
lund5000, line 105: In an earlier commit I was using an [mj-html-attributes](https://documentation.mjml.io/#mj-html-attributes) tag to put the `role` attribute on the enclosing `<div>`, but the build complained. MJML added that tag in 4.7.0, so maybe our library is a bit out of date?

apps/concierge_site/assets/mjml/notification.mjml
```diff
@@ -99,10 +99,10 @@
     <mj-raw>
       <!-- Footer -->
     </mj-raw>
-    <mj-section padding="0">
+    <mj-section padding="0" css-class="footer">
```
ErinLMoore, line 102: I'm not familiar with a `css-class` attribute, what does it do?

lund5000: It adds that class to the outermost element. This is actually a remnant of the mj-html-attributes approach I tried first, which relies on having a selector. I could remove this if we stick with adding the `role` to the `<p>` tag inside this section.

ErinLMoore: Oh, cool! Hadn't seen it before.

ErinLMoore (approved):
👍🏻

PR #1367, opened 2023-11-06
feat: refactor to use MBTA fork of `ueberauth_oidc`

paulswartz:
Updates performed on login-dev to support this:
- add a `roles` claim to the t-alerts client, configured to go in the `userinfo` output (`access_token` is not convertible to claims in the updated `ueberauth_oidc`)
- add a `sub` claim which maps to the `mbta_uuid` user attribute (I did this because it was easier to test in the sandbox where there's no `mbta_uuid`, but I can also revert that change and go back to using the `mbta_uuid` claim directly)

## to test
- [x] e-mail change functionality
- [x] admin functionality


arkadyan:
I think that https://github.com/mbta/alerts_concierge/pull/1368 should address the falling test case.

config/runtime.exs
```diff
@@ -21,24 +21,28 @@ config :alert_processor,
...
+  config :concierge_site,
+    keycloak_base_uri: base_uri
+
+  keycloak_base_opts = [
+    discovery_document_uri: System.fetch_env!("KEYCLOAK_WELL_KNOWN_OIDC"),
+    client_id: System.fetch_env!("KEYCLOAK_CLIENT_ID"),
+    client_secret: System.fetch_env!("KEYCLOAK_CLIENT_SECRET")
   ]
 
+  realm = System.get_env("KEYCLOAK_REALM") || "MBTA"
```
arkadyan, line 34: suggestion (non-blocking): There's also the 2-arity version of `System.get_env` that lets you provide the default.

arkadyan (approved):
Cool, this all feels much nicer, thanks!

config/runtime.exs
paulswartz: good point! I think I was thinking this also handled the `""` case, but it doesn't so better to use `get_env/2`

arkadyan:
@paulswartz I merged https://github.com/mbta/alerts_concierge/pull/1368 so feel free to rebase.

PR #1371, opened 2023-12-01
feat: switch to OIDC library that's actually released

paulswartz:
- Oidcc: https://hexdocs.pm/oidcc/readme.html
- UeberauthOidcc: https://hexdocs.pm/ueberauth_oidcc/readme.html

As far as I can tell, the test failures are only in `alerts_processor` and are also happening on `main`.


arkadyan:
That failing test smells like a GTFS change issue unrelated to this PR. I can check that out.

arkadyan:
https://github.com/mbta/alerts_concierge/pull/1372 should fix that test case.

arkadyan (approved):
Thanks, this is great. It took me a minute to spot the extra "c" and understand what I was looking at—I hadn't come across this library before. This is very timely as I'm just today adding SSO support to our new project, so I will shift direction and use this library.

paulswartz:
It's a fairly new library, but I wanted something that 1) supported log out fairly natively 2) was in Hex and ended up with Oidcc (even OpenID certified!) and the UeberauthOidcc wrapper around it.

arkadyan:
Ugh, the data changed again. I have a new fix.

PR #1375, opened 2023-12-06
fix: don't put the `logout_uri` in the Guardian claims

paulswartz:
Putting it there results in a very large session cookie, as we're nesting 3 JWTs (ID token JWT in the logout_uri -> Guardian JWT -> Session JWT) which can push things over the 4096 byte cookie limit.

### DIFF
--- a/apps/concierge_site/lib/controllers/auth_controller.ex
+++ b/apps/concierge_site/lib/controllers/auth_controller.ex
@@ -44,7 +44,9 @@ defmodule ConciergeSite.AuthController do
 
     {:ok, logout_uri} = UeberauthOidcc.initiate_logout_url(auth, logout_params)
 
-    SessionHelper.sign_in(conn, user, %{logout_uri: logout_uri})
+    conn
+    |> put_session("logout_uri", logout_uri)
+    |> SessionHelper.sign_in(user)
   end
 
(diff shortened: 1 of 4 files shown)


apps/concierge_site/lib/controllers/auth_controller.ex
```diff
@@ -44,7 +44,9 @@ defmodule ConciergeSite.AuthController do
 
     {:ok, logout_uri} = UeberauthOidcc.initiate_logout_url(auth, logout_params)
 
-    SessionHelper.sign_in(conn, user, %{logout_uri: logout_uri})
+    conn
+    |> put_session("logout_uri", logout_uri)
+    |> SessionHelper.sign_in(user)
```
arkadyan, line 49: suggestion (non-blocking): `SessionHelper.sign_in` probably doesn't need to accept the `claims` param any longer.

arkadyan (approved):
👍

apps/concierge_site/lib/controllers/auth_controller.ex
arkadyan: That is also a good point, probably not. We did have a React admin section back in the day, but Cora pulled that out a couple of years back.

PR #1382, opened 2024-01-22
Add urls to notifications

ErinLMoore:
Asana ticket: https://app.asana.com/0/1167196461144361/1206292714715118/f

- Adds notification urls to sms and plain text emails
- removes margin: 0 from emails


arkadyan (approved):
👍

PR #1386, opened 2024-02-06
Fix Typo

ErinLMoore:
well, the deploy went green, again. hopefully this does it.


arkadyan (changes requested):
This looks good, but when you first mentioned the issue the one I spotted was actually a different one in [Repo](https://github.com/mbta/alerts_concierge/blob/main/apps/alert_processor/lib/repo.ex#L13).

arkadyan (approved):
👍

PR #1391, opened 2024-02-28
add unsubscribe route

ErinLMoore:
Asana ticket: https://app.asana.com/0/1167196461144361/1206700151817926/f

### DIFF
--- a/apps/concierge_site/lib/controllers/unsubscribe_controller.ex
+++ b/apps/concierge_site/lib/controllers/unsubscribe_controller.ex
@@ -1,6 +1,6 @@
 defmodule ConciergeSite.UnsubscribeController do
   use ConciergeSite.Web, :controller
-  alias AlertProcessor.Model.Trip
+  alias AlertProcessor.Model.{Trip, User}
   @thirty_days_in_seconds 2_592_000
 
   def update(conn, %{"id" => encrypted_user_id}) do
@@ -13,6 +13,9 @@ defmodule ConciergeSite.UnsubscribeController do
 
     Trip.get_trips_by_user(user_id)
     |> Enum.each(&Trip.pause(&1, user_id))
+
+    user = User.get(user_id)
+    User.update_account(user, %{communication_mode: "none"}, user)
 
     json(conn, %{status: "ok"})
   end
--- a/apps/concierge_site/lib/controllers/unsubscribe_controller.ex
+++ b/apps/concierge_site/lib/controllers/unsubscribe_controller.ex
@@ -1,8 +1,16 @@
 defmodule ConciergeSite.UnsubscribeController do
   use ConciergeSite.Web, :controller
   alias AlertProcessor.Model.Trip
+  @thirty_days_in_seconds 2_592_000
 
-  def update(conn, %{"id" => user_id}) do
+  def update(conn, %{"id" => encrypted_user_id}) do
+    # We're not using a separate salt here because we're not
+    # encrypting passwords and don't need the extra security
+    {:ok, user_id} =
+      Plug.Crypto.decrypt(conn.secret_key_base, conn.secret_key_base, encrypted_user_id,
+        max_age: @thirty_days_in_seconds
+      )
+
     Trip.get_trips_by_user(user_id)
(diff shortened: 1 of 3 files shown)


apps/concierge_site/lib/controllers/unsubscribe_controller.ex
```diff
@@ -0,0 +1,19 @@
...
+    # encrypting passwords and don't need the extra security
+    {:ok, user_id} =
+      Plug.Crypto.decrypt(conn.secret_key_base, conn.secret_key_base, encrypted_user_id,
+        max_age: @thirty_days_in_seconds
+      )
+
+    Trip.get_trips_by_user(user_id)
+    |> Enum.each(&Trip.pause(&1, user_id))
+
+    json(conn, %{status: "ok"})
```
arkadyan, line 17: thought: Interesting that nothing seems to describe the structure of the expected response. (This seems as fine as anything.)

apps/concierge_site/test/web/controllers/unsubcribe_controller_test.exs
```diff
@@ -0,0 +1,66 @@
+defmodule ConciergeSite.UnsubscribeControllerTest do
+  @moduledoc false
+  use ConciergeSite.ConnCase, async: true
+  import AlertProcessor.Factory
+  alias AlertProcessor.Model.Subscription
+  alias AlertProcessor.Repo
+
+  describe "update/2" do
+    test "unsubscribes user from al subscriptions", %{conn: conn} do
```
arkadyan, line 9:
```suggestion
    test "unsubscribes user from all subscriptions", %{conn: conn} do
```

nitpick: Typo

apps/concierge_site/lib/controllers/unsubscribe_controller.ex
```diff
@@ -0,0 +1,19 @@
...
+  def update(conn, %{"id" => encrypted_user_id}) do
+    # We're not using a separate salt here because we're not
+    # encrypting passwords and don't need the extra security
+    {:ok, user_id} =
+      Plug.Crypto.decrypt(conn.secret_key_base, conn.secret_key_base, encrypted_user_id,
+        max_age: @thirty_days_in_seconds
+      )
+
+    Trip.get_trips_by_user(user_id)
+    |> Enum.each(&Trip.pause(&1, user_id))
```
arkadyan, line 15: suggestion: As a "belt and suspenders" approach, I wonder if it makes sense to also set the user's `communication_mode` to `none`.

apps/concierge_site/lib/controllers/unsubscribe_controller.ex
```diff
@@ -0,0 +1,19 @@
+defmodule ConciergeSite.UnsubscribeController do
+  use ConciergeSite.Web, :controller
+  alias AlertProcessor.Model.Trip
+  @thirty_days_in_seconds 2_592_000
+
+  def update(conn, %{"id" => encrypted_user_id}) do
+    # We're not using a separate salt here because we're not
+    # encrypting passwords and don't need the extra security
+    {:ok, user_id} =
+      Plug.Crypto.decrypt(conn.secret_key_base, conn.secret_key_base, encrypted_user_id,
```
arkadyan, line 10: question: In non-test environments, is this `secret_key_base` value being added to the `conn` automatically? It looks like it's also an application environment variable, but if it's part of the conn, too, that's fine, I'm just not familiar with it.

arkadyan (approved):
A small suggestion and question, but otherwise looks good.

apps/concierge_site/lib/controllers/unsubscribe_controller.ex
ErinLMoore: It apparently does read the secret_key_base out of the config (if available) and puts it in the conn automatically, so I could get it from either. I can't find anything about that in the documentation, but it works fine in my local dev environment.

apps/concierge_site/lib/controllers/unsubscribe_controller.ex
ErinLMoore: That's a good idea, I'll put that in.

PR #1436, opened 2024-04-26
re-add notifications to queue in case of throttling error

ErinLMoore:
Asana ticket: https://app.asana.com/0/1167196461144361/1207160402894161/f

This re-adds sms messages that came back with a 400 to the sending queue. Other errors are not added but there's additional logging so we can get a handle on how often it's happening. My main concern is a message somehow getting put in the queue over and over, although that doesn't really seem like a huge risk. If you think we should just retry all messages that come back with errors we can do that instead. (NB- as this is currently written, that would apply to emailed notifications too).


apps/alert_processor/lib/dissemination/notification_worker.ex
```diff
@@ -37,10 +37,32 @@ defmodule AlertProcessor.NotificationWorker do
...
-        send(self(), :work)
+        case NotificationSender.send(notification) do
+          {:ok, _} ->
+            log(id, notification, "event=send notification=#{id} time=#{now() - send_start}")
+            send(self(), :work)
+
+          {:error, {:http_error, 400, _}} ->
+            SendingQueue.push(notification)
+
+            log(
```
arkadyan, line 49: suggestion (non-blocking): I don't have strong feelings on this, so take it or leave it, but I wonder if this log line should be a warning and the one below should be an error?

arkadyan: As I think about it, the error one seems like a strong case to me since it's a failure that we're not able to properly recover from.

arkadyan (approved):
👍 I agree with you that I think this is the better approach to take.

apps/alert_processor/lib/dissemination/notification_worker.ex
ErinLMoore: Good point, will do.

PR #1440, opened 2024-04-30
fix: Remove unused claims argument on sign_in

arkadyan:
Asana ticket: [T-Alerts auth cleanup](https://app.asana.com/0/1204819229687089/1206753793118218)

### DIFF
--- a/apps/concierge_site/lib/helpers/session_helper.ex
+++ b/apps/concierge_site/lib/helpers/session_helper.ex
@@ -11,9 +11,9 @@ defmodule ConciergeSite.SessionHelper do
 
   @doc "Signs in a user with Guardian and redirects to the appropriate route."
   @spec sign_in(Conn.t(), User.t()) :: Conn.t()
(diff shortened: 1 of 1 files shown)


apps/concierge_site/lib/helpers/session_helper.ex
```diff
@@ -11,9 +11,9 @@ defmodule ConciergeSite.SessionHelper do
 
   @doc "Signs in a user with Guardian and redirects to the appropriate route."
   @spec sign_in(Conn.t(), User.t()) :: Conn.t()
-  def sign_in(conn, user, claims \\ %{}) do
+  def sign_in(conn, user) do
     conn
-    |> ConciergeSite.Guardian.Plug.sign_in(user, claims)
+    |> ConciergeSite.Guardian.Plug.sign_in(user)
```
arkadyan, line 16: The library already defaults to `%{}` for the `claims` argument so we don't need to supply it.

PR #1445, opened 2024-06-03
Monthly New Users query update

ErinLMoore:
[Asana Ticket](https://app.asana.com/0/1204819229687089/1207463067725489/f)

This is a query that can be run from the admin panel. It was running quite slow but it turns only we only really need this data from the beginning of the previous month. This changes the start date of the query to be that.


ErinLMoore:
> Are there descriptions/docs/text/somethings in the admin panel that should be updated for this?
> 
> Approving because I don't think that's a blocker 👍

There isn't any existing documentation like that, no.

PR #1453, opened 2024-07-18
updates user_update_worker to take either id or mbta_uuid as user id

ErinLMoore:
[Asana Ticket](https://app.asana.com/0/1204819229687089/1207720143191506/f)

Updates the user update worker so it works whether the user has an id or mbta_uuid field.
 Updates auth_controller to not throw error if mbta_uuid is missing.

### DIFF
--- a/apps/alert_processor/lib/user_update_worker.ex
+++ b/apps/alert_processor/lib/user_update_worker.ex
@@ -169,15 +169,13 @@ defmodule AlertProcessor.UserUpdateWorker do
   defp update_user_record(%{id: user_id, mbta_uuid: mbta_uuid, updates: updates} = attribute_map) do
     user_list = User.get_by_alternate_id(attribute_map |> Map.delete(:updates))
 
-    case length(user_list) do
-      0 ->
+    case user_list do
+      [] ->
         # Ignore updates for a user that has not yet logged in to T-Alerts
         :ok
 
-      1 ->
+      [user] ->
         # If 1 user is found, we want to update that user
-        user = hd(user_list)
-
         updates =
           Map.filter(
             %{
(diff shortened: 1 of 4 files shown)


apps/alert_processor/lib/user_update_worker.ex
```diff
@@ -185,6 +199,11 @@ defmodule AlertProcessor.UserUpdateWorker do
 
             :error
         end
+
+      2 ->
+        # If 2 users are found, something weird happened. Log and return nil.
+        Logger.warn("User with 2 ids found. sub id: #{user_id}, mbta_uuid: #{mbta_uuid}")
+        nil
```
arkadyan, line 206: suggestion: Should this be `:error` to match the function typespec?

arkadyan (approved):
One small thing, but otherwise looks good.

apps/alert_processor/lib/user_update_worker.ex
```diff
@@ -157,13 +166,18 @@ defmodule AlertProcessor.UserUpdateWorker do
   # Ignore empty updates. The user changed a property we aren't interested in saving locally.
   defp update_user_record(%{updates: updates}) when updates == %{}, do: :ok
 
-  defp update_user_record(%{user_id: user_id, updates: updates}) do
-    case User.get(user_id) do
-      nil ->
+  defp update_user_record(%{id: user_id, mbta_uuid: mbta_uuid, updates: updates} = attribute_map) do
+    user_list = User.get_by_alternate_id(attribute_map |> Map.delete(:updates))
+
+    case length(user_list) do
```
paulswartz, line 172: suggestion: rather than calculating the length of the list (which is linear in the length of the list) and matching on that, you can match on the list directly:
```elixir
case user_list do
  [] -> # empty case
  [user] -> # single user case
  _ -> # more than 1 user case
end
```

arkadyan: 💡

ErinLMoore: 💡

PR #1455, opened 2024-10-04
Add secure headers

ErinLMoore:
[ticket](https://app.asana.com/0/1204819229687089/1208252819442811/f)

A recent security audit revealed we need to add some more secure headers. The `put secure browser headers` plug takes care of most of them, so we were really only missing 2.


apps/alert_processor/test/alert_processor/service_info/service_info_cache_test.exs
```diff
@@ -248,7 +248,6 @@ defmodule AlertProcessor.ServiceInfoCacheTest do
                stop_list: [
                  {"Hingham", "Boat-Hingham", {42.253956, -70.919844}, 1},
                  {"Rowes Wharf", "Boat-Rowes", {42.355721, -71.049897}, 1},
-                 {"Georges Island", "Boat-George", {42.319742, -70.930427}, 1},
```
sloanelybutsurely, line 251: rip

PR #1458, opened 2025-03-11
feat: implement admin dashboard for South Coast Rail migration

boringcactus:
Some highlights:
- Since the subscriptions have UUIDs, if we derive a UUID for the copied subscription deterministically from the initial subscription, we can then check if we've already copied that subscription and skip it, making the migration phase 1 idempotent. (The route on a subscription can't be changed, as far as I can tell, so we don't have to worry about creating a duplicate-of-a-duplicate 16 times and looping back around to the original UUID again.)
- I'm not actually certain whether or not creating duplicate `NotificationSubscription`s is necessary to avoid resending notifications for active alerts at migration time, but it seemed like it might be.
- Since we've got Bootstrap 4 already set up, I'm using a [Bootstrap modal](https://getbootstrap.com/docs/4.6/components/modal/) to confirm the destructive migration phase 2.
- I tried fitting the entire migration into the existing admin queries interface, but copying the associated `NotificationSubscription`s was hard to do in pure SQL (and the UUID rewriting was less straightforward, and that code seems like it's only designed for read-only queries anyway).

### DIFF
--- a/apps/alert_processor/lib/model/notification_subscription.ex
+++ b/apps/alert_processor/lib/model/notification_subscription.ex
@@ -4,6 +4,8 @@ defmodule AlertProcessor.Model.NotificationSubscription do
   """
   use Ecto.Schema
 
+  @type t :: %__MODULE__{}
+
   schema "notification_subscriptions" do
     belongs_to(:notification, AlertProcessor.Model.Notification, type: :binary_id)
     belongs_to(:subscription, AlertProcessor.Model.Subscription, type: :binary_id)
--- a/apps/concierge_site/lib/controllers/admin/scr_controller.ex
+++ b/apps/concierge_site/lib/controllers/admin/scr_controller.ex
@@ -1,0 +1,214 @@
+defmodule ConciergeSite.Admin.ScrController do
+  use ConciergeSite.Web, :controller
+  import Ecto.Query
+  alias AlertProcessor.Model.NotificationSubscription
+  alias AlertProcessor.Model.Subscription
+  alias AlertProcessor.Repo
+
+  @old_route "CR-Middleborough"
+  @new_route "CR-NewBedford"
+  @old_stop_id "place-MM-0356"
+  @new_stop_id "place-MBS-0350"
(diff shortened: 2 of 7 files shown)


ErinLMoore:
I pulled down this branch to run it locally. The phase 1 button does duplicate the subscription, but when I go back to look at it in the UI, I get this kind of amalgam subscription, I assume because both the old and new subs have the same trip_ids. If I go to edit it, I get a 500, I think because the new route doesn't have any schedules yet.

Screenshot 2025-03-12 at 9 33 37 AM

If I go in and duplicate the trip, give it a new ID, and update the new New Bedford sub to have that trip_id, the subscriptions do appear separately in the UI, although the new one still throws an error upon editing. Do we want to duplicate the trips too so that the old subs are still editable (we'd have to remember to delete them along with the old subs in phase 2)?  On the other hand if they do make edits they'll be lost when we delete the old subs, maybe it doesn't matter.

boringcactus:
It’s weird that you’re getting a 500; when I tried to edit a trip in that state, it loaded fine but wouldn't save because there were no CR-NewBedford trips to select.

arkadyan:
@ErinLMoore Is it possible you might be running into the caching we have for dev and test to speed things up locally? It might be worth wiping it out with `rm apps/alert_processor/priv/service_info_cache/dev_cache.terms` (or the equivalent for test) to force it to re-download the latest schedule data.

(On a related note, it's most certainly a huge failing that this doesn't appear to be documented anywhere…)

arkadyan:
As I reread your comments, maybe my comment is a red herring and this has nothing to do with the cache. If so, I apologize for the noise.

ErinLMoore:
Ah, no, Matt's right, it's totally the dev cache. I cleared it and it works fine. Didn't realize it could cause this type of behavior. It still does combine the subscriptions but that's probably a good thing and not something we want to prevent. I rescind my comment 😅

ErinLMoore (approved):
This looks really good, thanks for taking this on.

PR #1459, opened 2025-03-12
docs: describe the data model

boringcactus:
This is worth preserving somewhere, and this seems like the most discoverable place to put it.

I'm fairly certain that this is accurate, but my experience with this app is very limited; it's possible that there are errors here or other things that would be worth mentioning.

### DIFF
--- a/apps/concierge_site/lib/controllers/admin/scr_controller.ex
+++ b/apps/concierge_site/lib/controllers/admin/scr_controller.ex
@@ -1,7 +1,6 @@
 defmodule ConciergeSite.Admin.ScrController do
   use ConciergeSite.Web, :controller
   import Ecto.Query
-  alias AlertProcessor.Model.NotificationSubscription
   alias AlertProcessor.Model.Subscription
   alias AlertProcessor.Repo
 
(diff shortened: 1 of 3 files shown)


docs/data-model.md
```diff
@@ -0,0 +1,29 @@
+# alerts_concierge Data Model
+
+## User-Driven
+
+In alerts_concierge, *trips* (referred to in the UI as “subscriptions”) represent an entire commute, potentially including transfers and a return journey.
+
+A trip is composed of multiple *subscriptions*, representing an individual leg of a commute; subscriptions have a single time range, a single route, and a single pair of stops where they start and end.
```
arkadyan, line 7: note: If I remember correctly, bus subscriptions work slightly differently and do not include stop and end stops. (I think the reasoning is that we don't need to ask the rider for more than that since, in practice, bus alerts are only created for entire routes.)

There's also a thing where you can have "parallel" bus routes in your trip. I personally use this: when I'm traveling to the office, I can take either the 131 or the 137 to get to the Orange Line, so I put both into my subscription and get alerts for either route.

docs/data-model.md
```diff
@@ -0,0 +1,29 @@
...
+Transfers create different subscriptions on the same trip (with a higher `rank`), the Green Line trunk creates different subscriptions on the same trip (one per `route`), and return trips create different subscriptions on the same trip (with `return_trip` `true` and the opposite `direction_id`).
+
+For commuter rail and ferry, to support riders who plan around individual scheduled trips, the UI presents a list of trips during the selected time range, appearing to allow the individual scheduled trips themselves to be included or excluded from the subscription.
+However, the subscription only tracks this information in `travel_start_time`/`travel_end_time`, which is set in the interface to between the first and the last selected trip, and indeed the interface does not allow the selection of non-contiguous sets of trips.
+Storing selected trips by ID would lead to selections losing meaning at new ratings if trip IDs change; this approach allows an equivalent trip at the same time to remain selected if its ID changes in the new rating.
```
arkadyan, line 12: note: Yes, and this is more of a "when" than an "if". CR trip IDs change every rating (or at least can commonly?). Usually, though, a trip is changing by a minute or two here or there, so as you say, the travel start and end times are a hack to try to represent "this trip that exists today and the 'essentially same' trip that will replace it next quarter, etc."

docs/data-model.md
```diff
@@ -0,0 +1,29 @@
...
+
+A trip can instead represent a non-time-windowed interest in accessibility information (`trip_type` of `accessibility` rather than `commute`, and subscription `type`s of `accessibility` rather than `subway`/`bus`/`cr`/`ferry`).
```
arkadyan, line 14: question: Can a "travel" trip also include accessibility subscriptions? I know we present it that way in the UI, but can't remember if we separate out multiple data structures on the back end.

arkadyan (commented):
This is excellent, thank you! I threw in a couple of drive by comments based on recollections, but I haven't actually validated anything, so take them with a grain of salt… I also went through this quickly and would definitely not consider my responses comprehensive. I'm sure Erin will do a much better job.

docs/data-model.md
```diff
@@ -0,0 +1,29 @@
+# alerts_concierge Data Model
```
arkadyan, line 1: suggestion: Maybe it's worth linking to this from the README for added visibility?

docs/data-model.md
boringcactus: Ah, yeah, good catch. I'm not sure if that reasoning is quite accurate, though - bus detours, for example, can specify a range of stops that may or may not include the segment a rider cares about.

docs/data-model.md
boringcactus: Added context.

docs/data-model.md
boringcactus: Yes, it can. Rephrased this paragraph.

docs/data-model.md
boringcactus: Linked.

docs/data-model.md
arkadyan: That's true, good point. Maybe it was more about making it less cumbersome for riders to get through the subscription process. It was a long time ago and my memory isn't always great… (But if either of you are ever interested, Jessie R was the designer on the project and might also have some valuable recollections.)

PR #1461, opened 2025-03-17
fix: raise timeout on some big SCR migration queries

boringcactus:
If this isn't sufficient, then we might be able to partition out the subscription UUIDs by leading digit and migrate 1/16 of them at a time, but I hope we don't need to do that. If this app already had Oban set up I'd be just running this as an Oban job instead, but setting that up just for this feels like more trouble than it's worth.


ErinLMoore (approved):
Let's give it a shot

PR #1462, opened 2025-03-19
fix: skip notification_subscriptions in SCR migration

boringcactus:
Since these take too long to load, and the risk created by not copying them is not severe, this seems like the best approach.


ErinLMoore (approved):
✅ looks good, let's hope it works!

PR #1463, opened 2025-05-23
feat: Add MBTA Go call to action for bus delay alerts

EmmaSimon:
Add a new line of text at the end of any bus delay alerts, letting users know they can check dotcom or the app for realtime information.

This adds a new column to the notifications table in the DB to keep a record of which CTA was included in their notification.

We include an `https://` scheme in the URL so that email clients render them as links, but remove them in texts, to save characters and because text clients don't usually require the scheme to render it as a link.

⚠️  This also shouldn't be merged before the /route redirects are deployed to dotcom prod and the MBTA Go URL is finalized, we'll be switching it to a deep link that directly opens the app, likely go.mbta.com

Screenshot_20250523-173305 Screenshot 2025-05-23 at 5 35 45 PM

## Testing

Added a few new unit tests to check the added behavior for notifications and SMS formatting, but I couldn't find any existing email formatting tests to add to.

There are currently 3 failing tests on main, I fixed one of them in this PR, but I'm not sure what's wrong with the other ones.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1209113360459249

### DIFF
--- a/apps/alert_processor/test/alert_processor/rules_engine/notification_builder_test.exs
+++ b/apps/alert_processor/test/alert_processor/rules_engine/notification_builder_test.exs
@@ -115,7 +115,7 @@ defmodule AlertProcessor.NotificationBuilderTest do
 
       assert %Notification{
                call_to_action:
-                 "Track your bus at https://mbta.com/route/1 or use the MBTA Go app: https://mbta.com/mobile-app"
+                 "Track your bus at https://mbta.com/route/1 or use the MBTA Go app: https://go.mbta.com"
              } = notification
     end
 
@@ -141,8 +141,7 @@ defmodule AlertProcessor.NotificationBuilderTest do
       notification = NotificationBuilder.build_notification(user_subs, alert)
 
       assert %Notification{
-               call_to_action:
-                 "To track your bus, use the MBTA Go app: https://mbta.com/mobile-app"
+               call_to_action: "To track your bus, use the MBTA Go app: https://go.mbta.com"
              } = notification
     end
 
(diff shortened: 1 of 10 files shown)


apps/alert_processor/test/alert_processor/rules_engine/notification_builder_test.exs
```diff
@@ -83,4 +85,84 @@ defmodule AlertProcessor.NotificationBuilderTest do
...
+      }
+
+      notification = NotificationBuilder.build_notification(user_subs, alert)
+
+      assert %Notification{
+               call_to_action: "To track your bus, use the MBTA Go app: https://go.mbta.com"
+             } = notification
+    end
+
+    test "nil cta when not a delay alert", %{user_subs: user_subs} do
```
ErinLMoore, line 148: I'd probably add a quick test making sure it behaves correctly for alerts where the effect name is "Delay" but the mode is something else- eg Commuter Rail delay alerts where the route_type is 2.

apps/alert_processor/test/alert_processor/rules_engine/notification_builder_test.exs
```diff
@@ -83,4 +85,84 @@ defmodule AlertProcessor.NotificationBuilderTest do
...
+      sub = insert(:subscription, user: user)
+
+      {:ok, user_subs: {user, [sub]}}
+    end
+
+    test "includes dotcom url and mbta go CTA if single bus route is effected", %{
+      user_subs: user_subs
+    } do
+      now = naive_to_local(~N[2018-01-11 14:10:55Z])
+      one_hour_ago = DateTime.add(now, -3600)
```
ErinLMoore, line 101: (second syntax nitpick) This is just personal preference, but this can also be written DateTime.add(now, -1, :hour)- you can also do DateTime.add(now, 1, :day) for days and so forth.

ErinLMoore (approved):
This looks good, I added a comment about a test just to cover one of the code branches, but I think it'll work fine.

PR #1467, opened 2025-12-15
fix: Log as warning for sending failures that get retried

arkadyan:
These 400 failures typically represent throttling responses from AWS. We
are able to retry sending, which generally results in success. Given
this, a warning log message is sufficient. This prevents us from
triggering an alert due to an expected situation when we are sending a
high volume of notifications in a short time period.


arkadyan:
@paulswartz when you have a spare moment, could you please look this over? This is a small improvement I came up with based on some recent experiences with false incident alarms.

paulswartz (approved):

PR #1474, opened 2026-06-11
fix: Support translated effect_detail strings

EmmaSimon:
Ticket: [T-Alerts | Support translated string for `effect_detail`](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215199208415401?focus=true)

Alerts Manager is planning to update the GTFS-RT enhanced JSON feed to more closely match the spec, by translating the `effect_detail` and `cause_detail` fields. Alerts Concierge doesn't use the RT feed's `cause_detail` field anywhere, so we don't need to worry about parsing that, but it does use the `effect_detail`. This handles the change by accepting either a basic string like it currently sends, or by parsing a translated string out of an object like some of the other fields are already doing.

I updated some of the test data to be in the new translation object format, and tests still pass. Also added some unit tests for new parse function.


apps/alert_processor/lib/alert_parsing/alert_parser.ex
```diff
@@ -210,6 +210,18 @@ defmodule AlertProcessor.AlertParser do
 
   defp parse_duration_certainty(alert, _, _, _), do: %{alert | duration_certainty: :known}
 
+  @spec parse_effect_name(map | list | String.t()) :: String.t() | nil
+  def parse_effect_name(translations), do: do_parse_effect_name(translations)
```
KaylaBrady, line 214: question(non-blocking): why do we need separate `parse_effect_name` and `do_parse_effect_name` functions?

KaylaBrady (approved):

apps/alert_processor/lib/alert_parsing/alert_parser.ex
EmmaSimon: I was just following the pattern of the other helpers here. I think it's partly for testing, the `parse_*` functions are public and the `do_parse_*` ones are private.

